PR #9, opened 2019-01-11
[DNM] upmerge to upsteram 9f1933

nvlsianpu:
This is fast forward as NCS doesn't apply internal changes so far.

This is replay of #7.

New version introduces:

- fix for image encryption feature
- Image copy procedure would try to decode image while enc-key is not provided.
- RAM usage reduction by a few bytes

- Remove unused hal includes from bootutil

- Fix encrypted support for images with header sizes larger than 256 bytes

- Multiple flash device support capability

- Add boostrapping from slot1 support

- Intel S1000 support (xtensa board)

- AES key unwrapping following RFC-3394 section 2.2.2 using tinycrypt for AES-128 decryption. It is enabled when ECDSA was chosen as the signature algorithm. (The key unwrap functionality was already provided only using mbedTLS)

- Update Mbed-TLS submodule to 2.14.1
- Update bundled ASN1 parser to Mbed-TLS 2.14.1
- Enable ASN1 parsing for KW

- Fix overwrite only mode (#if-def-ry) for zephyr builds

- imgtool save to format of given output filename

- switch logging to LOG


nvlsianpu:
@carlescufi CLA verification is theoretically off, but has pending on this PR). Although seams not produced any results. Let's ignore it here.

nvlsianpu:
@carlescufi  Can you look at this?

nvlsianpu:
Ups - I would merge this just using push. This is still just upsteram mirror.

carlescufi:
Right, sorry! Since there is no merge commit it makes no difference in this case right? can you verify we are still identical to upstream?

nvlsianpu:
We are identical.

PR #13, opened 2019-02-19
[nrf noup] ci: Applied fix for compliance script parameters

ChristopherBittner:
Changed of the way that parameters are set for the compliance script. Previously that was failing for branches that would not require additional report parameters to be send for the script. This PR should fix false negative results of the compliance script for master and the branches.

Only pull requests will receive parameters for the comparison of commits. 
Otherwise will compare with the origin.

Signed-off-by: Chris Bittner <chris.bittner@nordicsemi.no>


ChristopherBittner:
cc @nvlsianpu

PR #16, opened 2019-03-20
[nrf fromlist] cmake: kconfig: zephyr multi image built support

SebastianBoe:
dump

Signed-off-by: Sebastian Bøe <sebastian.boe@nordicsemi.no>

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -91,14 +91,6 @@ endchoice
 endif
 endchoice
 
-config BOOT_SIGNATURE_KEY_FILE
-	string "PEM key file"
-	default ""
-	help
-	  The key file will be parsed by imgtool's getpub command and a .c source
-	  with the public key information will be written in a format expected by
-	  MCUboot.
-
 config MBEDTLS_CFG_FILE
 	default "mcuboot-mbedtls-cfg.h"
 
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -1,0 +1,39 @@
+if(CONFIG_BOOTLOADER_MCUBOOT)
+  # Build a second bootloader image
+
+  set(MCUBOOT_BASE ${CMAKE_CURRENT_LIST_DIR}/..)
+
+  zephyr_add_executable(mcuboot require_build)
+
+  if (${require_build})
+    add_subdirectory(${MCUBOOT_BASE}/boot/zephyr ${CMAKE_CURRENT_BINARY_DIR}/mcuboot)
+
+    # TODO: Assert that the bootloader and image use the same key.
+
+    set(SIGNED_IMAGE signed.hex)
+
+    set_property(GLOBAL APPEND PROPERTY
+      extra_post_build_commands
+      COMMAND
+      ${PYTHON_EXECUTABLE}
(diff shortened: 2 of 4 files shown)


nvlsianpu:
should I merge this? XD

hakonfam:
Hehe, @SebastianBoe  just went on holiday, and we had to get it in before he left. Hence the lack of a proper title and description for this PR. 

I updated the commit message, so that should be descriptive enough for the sake of this PR.

@nvlsianpu : Do you need me to open a new PR with a proper name, or can this one suffice? 

The PR needs to be synced with 
https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/567 
and
https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/110

boot/zephyr/Kconfig
```diff
@@ -60,14 +60,6 @@ config BOOT_SIGNATURE_TYPE_ECDSA_P256
 
 endchoice
 
-config BOOT_SIGNATURE_KEY_FILE
```
nvlsianpu: Why this property can't be there?
Removal will impose possible problems with any future upmerge (it is error prone).
I think dependency enlargement to  `MCUBOOT || BOOTLOADER_MCUBOOT` can go upstream.

hakonfam: It needs to visible for both the application including MCUboot, and MCUboot itself.
This because the application will use this configuration to invoke the `imgtool.py` script with the correct key file.

When there is support for sharing specific Kconfig options between images it would be possible to have this config here.

Could you explain why moving this option can become an issue?

nvlsianpu: I was not aware of counterpart PR to zephyr upstream at the moment I had ben asking.
Will be still possible to build as stan-alone project. I mean to build only mcuboot as we use to do so fare.

hakonfam: My bad, the details about the counterpart PR should have been stated more clearly.

The "normal" way of building MCUboot is not affected by this.

zephyr/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
...
+    set_property(GLOBAL APPEND PROPERTY
+      extra_post_build_commands
+      COMMAND
+      ${PYTHON_EXECUTABLE}
+      ${MCUBOOT_BASE}/scripts/imgtool.py
+      sign
+      --key ${MCUBOOT_BASE}/${CONFIG_BOOT_SIGNATURE_KEY_FILE}
+      --header-size ${CONFIG_TEXT_SECTION_OFFSET}
+      --align       ${DT_FLASH_WRITE_BLOCK_SIZE}
+      --version 0.1       # TODO: Configurable?
```
nvlsianpu: Do todo's around are supposed to be fulfilled in another PR?

hakonfam: Yes, because there is need for sharing configurations between multiple images, this is not supported currently. 

This is one of two places where this functionality is needed for now, and both of them are in MCUBoot.

hakonfam:
@nvlsianpu FYI the last commit removed a check surrounding a Kconfig which caused a CI error. This beause the Kconfig symbol being checked was declared in `boot/zephyr/Kconfig`, which is not parsed by the Kconfig used for CI check.

PR #20, opened 2019-04-02
zephyr: introduce partition manager

hakonfam:
Partition Manager is a component which uses yaml files to
resolve flash placement with a wholistic view of the
device

TODO
- Support skipping hex files. To support this we need to support having a pm.yml in the app folder which defines what sections should be assumed in place when executing partition_manager.py.
- Check PM_ variable in cmake to generate merged hex file


boot/zephyr/pm.yml
```diff
@@ -0,0 +1,12 @@
+mcuboot:
+  placement:
+    before: [spm, app]
+mcuboot_partitions:
+  inside: [spm, app]
```
lemrey: What does `inside` do?

boot/zephyr/Kconfig
```diff
@@ -14,6 +14,27 @@ config MCUBOOT
...
+
+
+config PARTITION_MANAGER_RESERVED_SPACE_MCUBOOT_SCRATCH
+	hex "Flash space reserved for scratch."
+	default 0x1e000
+	help
+	  Flash space set aside for the scratch area.
+
+config PARTITION_MANAGER_RESERVED_SPACE_MCUBOOT_STORAGE
+	hex "Flash space reserved for scratch."
```
lemrey: s/scratch/storage
In the help string too

boot/zephyr/pm.yml
hakonfam: I creates sub-partitions.  Note that the documentation will be done ASAP after code is in (hopefully that is OK).

For the case of MCUboot, it needs to have a concept of a partition which consists of everything that should be updated. This because it doesn't support multi image updates ATM.

When an app is built with SPM, those two images needs to be treated as one by mcuboot. This is what is represented here.

lemrey: If `inside` creates subpartitions, what does `subpartition` do?

boot/zephyr/Kconfig
```diff
@@ -14,6 +14,27 @@ config MCUBOOT
 	select MPU_ALLOW_FLASH_WRITE if ARM_MPU
 	select USE_CODE_PARTITION if HAS_FLASH_LOAD_OFFSET
 
+config PARTITION_MANAGER_RESERVED_SPACE_MCUBOOT
```
lemrey: Man, I wish these macros where shorter :) Somewhere else I have seen Partition Manager macros prefixed with just `PM`, that's maybe a bit cryptic; something in-between would be better (just on the top of my head: `PARTMAN` ??). 

`RESERVED_SPACE` is informative but `AREA` would also work and is the same terminology used in `flash_map` and Device Tree. Otherwise it might make sense to stick to your own terminology and call that a `PARTITION_SIZE`, same length.

In any case the macro prefix should be aligned.

boot/zephyr/include/sysflash/sysflash.h
```diff
@@ -3,10 +3,21 @@
 #ifndef __SYSFLASH_H__
 #define __SYSFLASH_H__
 
+#ifdef USE_PARTITION_MANAGER
+#include <pm_config.h>
+
+#define FLASH_AREA_IMAGE_PRIMARY    PM_CFG_MCUBOOT_PARTITIONS_PRIMARY_ID
```
lemrey: I would omit the `_CFG_` infix; don't find it very meaningful.

boot/zephyr/Kconfig
hakonfam: `PARTMAN` sounds awesome. A small comment could be enough to make it clear even with `PM` though.

boot/zephyr/include/sysflash/sysflash.h
hakonfam: will remove

boot/zephyr/Kconfig
hakonfam: Will change to `PM_PARTITION_SIZE_x` and add comments in the kconfig files

boot/zephyr/pm.yml
hakonfam: One level nesting is supported.

The second nested level only supports equal split of the parent partition.

It is done like this to support the MCUboot use case.

`inside` indicates where the listed `sub-partitions` will be placed.

hakonfam: I updated the keyword to be `span` instead. See if that makes more sense.

lemrey: `Define a set of sub-partitions which spans over 'mcuboot_pad', 'spm' and 'app'`
Does that mean that `primary` and `secondary` partitions have `mcuboot_pad`, `spm` and `app` "subpartitions"?

nvlsianpu:
Shouldn't mcuboot code partition, mcuboot scartch partition be unique in a device wholistic view? Can't it be extracted from DTS?

I think I don't understand well objective behind this patch. Can you elaborate on that?

hakonfam:
The partitions are unique, the problem lies with extracting this information from DTS.

Where mcuboot code partition and mcuboot scratch partition depends on Kconfig in a multi image context. An application might configure in images which is placed before mcuboot (imagine first stage immutable bootloader with capability to update mcuboot version). Also an application might include a sub-image which is placed after the app. Both of these cases require mcuboot partitions to be re-located based on kconfig options. This can not be achieved using DTS.

boot/zephyr/include/sysflash/sysflash.h
```diff
@@ -3,10 +3,21 @@
 #ifndef __SYSFLASH_H__
 #define __SYSFLASH_H__
 
+#ifdef USE_PARTITION_MANAGER
+#include <pm_config.h>
+
+#define FLASH_AREA_IMAGE_PRIMARY    PM_MCUBOOT_PARTITIONS_PRIMARY_ID
+#define FLASH_AREA_IMAGE_SECONDARY  PM_MCUBOOT_PARTITIONS_SECONDARY_ID
+#define FLASH_AREA_IMAGE_SCRATCH    PM_MCUBOOT_SCRATCH_ID
```
SebastianBoe: Should this be MCUBoot-specific?

hakonfam: Do you mean the partitions? Or the code in this file?

No, ideally it should be shared between zephyr and mcuboot. I just had a bug because of this split :) Not sure if changing it is in the scope of this PR-set though.

boot/zephyr/pm.yml
```diff
@@ -0,0 +1,33 @@
...
+  # will only span across that partition.
+  # The sub partitions share the size of the parent partition(s) equally.
+  sub_partitions: [primary, secondary]
+  span: [mcuboot_pad, spm, app]
+
+# The size of this partition is defined by the kconfig symbol
+# CONFIG_PM_PARTITION_SIZE_MCUBOOT_SCRATCH
+mcuboot_scratch:
+  placement:
+    after: [app]
```
SebastianBoe: For my own understanding. Would it work to express "after: [mcuboot_partitions]" ?

hakonfam: It should work. Have not tested this yet.

PR #21, opened 2019-04-04
DNM [nrf mergeup] merge upstream up to commit 288682c

nvlsianpu:
* do not override TEXT_SECTION_OFFSET (The nrf52840_pca10059 board no longer
overrides TEXT_SECTION_OFFSET but sets the correct FLASH_LOAD_OFFSET
instead)
* Fix bug https://github.com/JuulLabs-OSS/mcuboot/issues/399 where dts.overlay was not being appended.
* imgtool disallow using public RSA key for signing.


nvlsianpu:
cc @Qbicz

PR #22, opened 2019-04-08
[DNM][nrf fromlist] cmake: Use slot size from DTS for signing

Qbicz:
Use maximal slot size as configured in DTS for signing with imgtool.py.

Jira: DESK-517

Signed-off-by: Filip Kubicz <filip.kubicz@nordicsemi.no>

### DIFF
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -12,6 +12,12 @@ if(CONFIG_BOOTLOADER_MCUBOOT)
 
     set(SIGNED_IMAGE signed.hex)
 
+    if (${DT_FLASH_AREA_IMAGE_0_SIZE} GREATER ${DT_FLASH_AREA_IMAGE_1_SIZE})
+      set(SLOT_SIZE ${DT_FLASH_AREA_IMAGE_0_SIZE})
+    else()
+      set(SLOT_SIZE ${DT_FLASH_AREA_IMAGE_1_SIZE})
+    endif()
+
     set_property(GLOBAL APPEND PROPERTY
       extra_post_build_commands
       COMMAND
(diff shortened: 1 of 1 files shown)


Qbicz:
This is to fix error when trying to build app bigger than 0x32000 bytes:
```
Image size + trailer exceeds requested size 0x32000
```

We could also push this value to 0x67000 (current slot value in DTS for nRF52840) and leave the TODO.

Qbicz:
@SebastianBoe @nvlsianpu can you review?

SebastianBoe:
@Qbicz : Hi, I'm sorry, but multi-image was merged pre-maturely. This and a host of other problems are fixed and will soon be merged/applied as a part of the partition manager PR's.

Fixing multi-image/mcuboot before the partition manager PR's are in is not desirable as the patches will only last a few days. But we should keep track of all the issues, so that we can verify they are all fixed after PM is merged.

@hakonfam : FYI

Qbicz:
@SebastianBoe thanks for clarifying. Partition manager is here: https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/612 together with zephyr and mcuboot PRs.

Can we revert/workaround the issues introduced by multi-image? Otherwise we will have broken master in NCS for a few days or over a week, depending on how long the review and merge process will take.

FYI @pdunaj @wiminordic

SebastianBoe:
AFAIK using multi-image is optional, so having it broken in-tree does not break anything.

Qbicz:
Actually, these are post-build steps that fail.

@pdunaj @MarekPieta, for now ignore failure occuring when building with MCUboot support. This is failure during signing, correct executables are there already. Instead, use our usual `ninja sign` and `ninja mcuboot`.

pdunaj:
@SebastianBoe I just tried to get our project to sign the image and it is not working. We have our commands to sign but this is not working as new signing step is always added (we depend on zephyr.elf and signing is performed while generating it). 
If this is optional can you tell how I can disable it? If this is not optional can the PR be accepted as we are unable to create a signed version of our binary.

SebastianBoe:
Which project is this? I would like to reproduce the issue.

SebastianBoe:
An MCUBoot-bootable image that wants to opt-out of using multi-image should use this configuration:

CONFIG_BOOTLOADER_MCUBOOT=y
CONFIG_MCUBOOT_BUILD_STRATEGY_SKIP_BUILD=y

nvlsianpu:
@SebastianBoe @Qbicz So conclusion for this PR i that it should be never merge. Em I right?

SebastianBoe:
Yes.

PR #23, opened 2019-04-08
cmake: check BOARD before setting default

hakonfam:
Currently the BOARD variable is not checked before setting the
default value. This causes issues in a multi image context.

Fix this by checking if the BOARD variable is set before
setting it to its default value.

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -8,7 +8,11 @@ cmake_minimum_required(VERSION 3.8.2)
 
 # Board-specific CONF_FILES should get merged into the build as well.
 # Default to qemu_x86 if no board has been specified.
-set(BOARD qemu_x86)
+# If BOARD is already defined, mcuboot is being built in a multi image
+# context, and should default to its parent image BOARD.
+if (NOT BOARD)
+  set(BOARD qemu_x86)
+endif()
 
 # Add a common dts overlay necessary to ensure mcuboot is linked into,
 # and fits inside, the boot partition. (If the user specified a


hakonfam:
> Could this go upstream as is?

Yes

pizi-nordic:
@ioannisg: IMHO it should go there.

SebastianBoe:
No, because in a non-multi image context it doesn't make sense to check if BOARD is set.

ioannisg:
> 
> 
> No, because in a non-multi image context it doesn't make sense to check if BOARD is set.

I see. Then, please, address my comment about making it more clear that this is only applicable when building multi-image. :)

hakonfam:
the PR upstream is opened here 

https://github.com/JuulLabs-OSS/mcuboot/pull/461

Due to tight timeline I was hoping to get this in downstream and then be out of sync until the next upmerge.

SebastianBoe:
> address my comment about making it more clear that this is only applicable when building multi-image.

We have a lot of code downstream that is only applicable when building multi-image. I don't think we want to wrap all of it with comments.

ioannisg:
> 
> 
> > No, because in a non-multi image context it doesn't make sense to check if BOARD is set.
> 
> I see. Then, please, address my comment about making it more clear that this is only applicable when building multi-image. :)

Also, please, explain the rationale in the comment that precedes the change.

ioannisg:
> 
> 
> > address my comment about making it more clear that this is only applicable when building multi-image.
> 
> We have a lot of code downstream that is only applicable when building multi-image. I don't think we want to wrap all of it with comments.

As far as I understood, this change is only needed when we build a Zephyr image as part of multi-image build. Is there a way to know "inside the build" that we are building Zephyr with multi-image support? A Kconfig or CMake setting, perhaps? If that is the case, I think it make sense to wrap the
```
if (NOT BOARD)
  set(BOARD qemu_x86)
endif()
```

with this conditional.
Adding a comment is optional, if the code itself is self-explanatory

hakonfam:
See the comment I added in the code @ioannisg

PR #26, opened 2019-04-12
zephyr: cmake: build update hex with --pad

hakonfam:
Prior to this, the update hex was deduced from the update.bin.
The problem is that update.bin does not have the end padding.
The end padding is required for mcuboot to upgrade a hex file
flashed directly to the secondary slot.

Fix this by having a separate sign command for update.hex which
includes the --pad argument. This enables the user to verify
the update process by flashing the update.hex file directly to
the board.


hakonfam:
Tested DFU with both update.bin from the web and flashing moved_update.hex using http_application_fota example

hakonfam:
https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/655

PR #32, opened 2019-04-29
[nrf noup] pm: Adapt to changes in partition manager and change names

oyvindronningstad:
Use 'inside' and 'inherit_size'.

Related PRs:
 - https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/138
 - https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/705


boot/zephyr/pm.yml
```diff
@@ -1,33 +1,38 @@
-# The size of this partition is defined by the kconfig symbol
-# CONFIG_PM_PARTITION_SIZE_MCUBOOT
+#include <autoconf.h>
```
hakonfam: We discussed automagically injecting these in the linker command, so that the .yml files does not have to have this line.

How do you feel about doing that? IMO it is easier to simply explain to the user that KCONFIG is available here, than to say that it is preprocessed AND that you can include any header file AND that this is the header file you want if you want kconfig options.

oyvindronningstad: I'm on the fence about it. The way shown here is more transparent. I get your point about explaining it, but if they are using Zephyr they are probably familiar with preprocessing, so it isn't a big jump. We anyway need to explain somewhere that it is preprocessed, or else they might get other mysterious surprises.

Also, auto-including needs a separate build-step, with an additional intermediate file, so this is simpler in that sense as well.

hakonfam: I agree, magic can be dangerous :)

oyvindronningstad:
@nvlsianpu @carlescufi This should be merged together with the referenced PRs, which will be ready shortly.

oyvindronningstad:
Ready to merge

PR #39, opened 2019-05-28
kconfig: add config for mcuboot version

hakonfam:
Make it easier for the user to change the mcuboot version
by extracting this as a kconfig variabl.

### DIFF
--- a/zephyr/Kconfig
+++ b/zephyr/Kconfig
@@ -33,6 +33,16 @@ config MCUBOOT_BUILD_STRATEGY_FROM_SOURCE
 
 endchoice
 
+config MCUBOOT_IMAGE_VERSION
+        string "Image version"
+        default "0.0.0+0"
+	help
+	  Value to be passed as 'version' argument to 'imgtool.py' when
+	  creating signed image. Note that no semantics are connected to
+	  this variable. It does not provide downgrade prevention, and is only
+	  valuable for debugging purposes. Format: maj.min.rev+build with
+	  latter parts optional.
+
 endif # BOOTLOADER_MCUBOOT
 
 config BOOT_SIGNATURE_KEY_FILE
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -37,8 +37,8 @@ if(CONFIG_BOOTLOADER_MCUBOOT)
     --key ${MCUBOOT_BASE}/${CONFIG_BOOT_SIGNATURE_KEY_FILE}
     --header-size $<TARGET_PROPERTY:partition_manager,MCUBOOT_HEADER_SIZE>
     --align       ${DT_FLASH_WRITE_BLOCK_SIZE}
-    --version 0.1 # TODO configurable?
-    --slot-size $<TARGET_PROPERTY:partition_manager,MCUBOOT_SLOT_SIZE>
+    --version     ${CONFIG_MCUBOOT_IMAGE_VERSION}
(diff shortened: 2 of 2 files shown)


zephyr/Kconfig
```diff
@@ -33,6 +33,16 @@ config MCUBOOT_BUILD_STRATEGY_FROM_SOURCE
 
 endchoice
 
+config MCUBOOT_VERSION
+        string "MCUbot version"
+        default "0.0.1"
```
SebastianBoe: Should match west's default, which is "0.0.0+0".

zephyr/Kconfig
```diff
@@ -33,6 +33,16 @@ config MCUBOOT_BUILD_STRATEGY_FROM_SOURCE
...
+	help
+	  Value to be passed as 'version' argument to 'imgtool.py'.
+	  Note that no semantics are connected to this variable.
```
SebastianBoe: > Note that no semantics are connected to this variable.

Is this likely to stay true? Should we say less here and reference external
documentation more?

zephyr/Kconfig
```diff
@@ -33,6 +33,16 @@ config MCUBOOT_BUILD_STRATEGY_FROM_SOURCE
 
 endchoice
 
+config MCUBOOT_VERSION
+        string "MCUbot version"
```
SebastianBoe:
```Suggestion
        string "MCUboot version"

File: zephyr/Kconfig [hunk 4]
```diff
@@ -33,6 +33,16 @@ config MCUBOOT_BUILD_STRATEGY_FROM_SOURCE
...
+        default "0.0.1"
+	help
+	  Value to be passed as 'version' argument to 'imgtool.py'.
+	  Note that no semantics are connected to this variable.
+	  It does not provide downgrade prevention, and is only valuable for
+	  debugging purposes. Format: maj.min.rev+build with latter parts
+	  optional.
```
SebastianBoe: Can we make it clearer that MCUBOOT_VERSION is not related to the MCUBoot version but the version of the image?

SebastianBoe: e.g. MCUBOOT_IMAGE_VERSION

zephyr/Kconfig
hakonfam: I am not aware of any documentation regarding this. Will take a second look.

hakonfam: I can't find any docs about this other than the code. 
It will not stay like this for long in our fork at least, but this is something we can ensure to keep updated without much hassle.

```
➜  mcuboot git:(kconfig-version) ✗ rg -t rst "version"
ext/tinycrypt/documentation/tinycrypt.rst
140:    be changed in future versions of the library as there are applications
➜  mcuboot git:(kconfig-version) ✗ 
```

zephyr/Kconfig
```diff
@@ -33,6 +33,16 @@ config MCUBOOT_BUILD_STRATEGY_FROM_SOURCE
 
 endchoice
 
+config MCUBOOT_IMAGE_VERSION
```
nvlsianpu: how does this corporate with what mcuboot/mcumgr can recognize (and logg to the console)?

nvlsianpu:
```
struct image_version {
    uint8_t iv_major;
    uint8_t iv_minor;
    uint16_t iv_revision;
    uint32_t iv_build_num;
};
```

PR #41, opened 2019-06-06
zephyr: cmake: Use PM properties with the proper names

oyvindronningstad:
This is needed when the properties are automatically set.

Signed-off-by: Øyvind Rønningstad <oyvind.ronningstad@nordicsemi.no>


nvlsianpu:
looks good, but commit need [nrf toup] tag

oyvindronningstad:
> sauce tag

Done

PR #50, opened 2019-08-28
Create update candidate for S1 and S0

hakonfam:
Sign both S0 and S1 (which is already signed by B0) in order to prepare update candidates for both slots.

### DIFF
--- a/zephyr/CMakeLists.txt
+++ b/zephyr/CMakeLists.txt
@@ -1,6 +1,75 @@
 if(CONFIG_BOOTLOADER_MCUBOOT)
-  # Build a second bootloader image
+  function(sign to_sign_hex output_prefix offset sign_depends signed_hex_out)
+    set(op ${output_prefix})
+    set(signed_hex ${op}_signed.hex)
+    set(${signed_hex_out} ${signed_hex} PARENT_SCOPE)
+    set(to_sign_bin ${op}_to_sign.bin)
+    set(update_bin ${op}_update.bin)
+    set(moved_test_update_hex ${op}_moved_test_update.hex)
+    set(test_update_hex ${op}_test_update.hex)
 
+    add_custom_command(
+      OUTPUT
+      ${update_bin}            # Signed binary of input hex.
+      ${signed_hex}            # Signed hex of input hex.
+      ${test_update_hex}       # Signed hex with IMAGE_MAGIC
+      ${moved_test_update_hex} # Signed hex with IMAGE_MAGIC located at secondary slot
+
+      COMMAND
+      # Create signed hex file from input hex file.
+      # This does not have the IMAGE_MAGIC at the end. So for this hex file
+      # to be applied by mcuboot, the application is required to write the
+      # IMAGE_MAGIC into the image trailer.
+      ${sign_cmd}
+      ${to_sign_hex}
+      ${signed_hex}
+
+      COMMAND
+      # Create binary version of the input hex file, this is done so that we
+      # can create a signed binary file which will be transferred in OTA
+      # updates.
+      ${CMAKE_OBJCOPY}
+      --input-target=ihex
+      --output-target=binary
+      ${to_sign_hex}
+      ${to_sign_bin}
+
+      COMMAND
+      # Sign the binary version of the input hex file.
+      ${sign_cmd}
+      ${to_sign_bin}
+      ${update_bin}
+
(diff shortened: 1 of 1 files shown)


hakonfam:
cc @oyvindronningstad @sigvartmh

zephyr/CMakeLists.txt
```diff
@@ -43,52 +106,73 @@ if(CONFIG_BOOTLOADER_MCUBOOT)
...
     mcuboot_primary_app_PM_TARGET
     mcuboot_sign_target
     )
+
+  if (CONFIG_MCUBOOT_BUILD_S1_VARIANT)
+    # Secure Boot (B0) is enabled, and we have to build update candidates
+    # for both S1 and S0.
+
+    foreach(slot s0_image;s1_image)
+      string(SUBSTRING ${slot} 0 2 slot_only) # 's0_image' -> 's0'
```
oyvindronningstad: How about `foreach(slot s0;s1)` instead? In that case, `${slot}` -> `${slot}_image` and `${slot_only}` -> `${slot}`.

zephyr/CMakeLists.txt
```diff
@@ -43,52 +106,73 @@ if(CONFIG_BOOTLOADER_MCUBOOT)
...
+      # Depend on both the target for the hex file, and the hex file itself.
+      set(dependencies "${${slot}_target};${${slot}_hex}")
+
+      set(out_path ${PROJECT_BINARY_DIR}/signed_by_mcuboot_and_b0_${slot})
+
+      sign(${${slot}_hex} # Hex file to sign
+        ${out_path}
+        ${slot_offset}
+        "${dependencies}"    # Need "..." to make it a list.
+        b_one_signed_hex     # Created file variable
```
oyvindronningstad: b_one == b1? IMO "signed_hex" is sufficient.

zephyr/CMakeLists.txt
```diff
@@ -1,6 +1,75 @@
 if(CONFIG_BOOTLOADER_MCUBOOT)
-  # Build a second bootloader image
+  function(sign to_sign_hex output_prefix offset sign_depends signed_hex_out)
```
oyvindronningstad: Could this function be upstreamed?

zephyr/CMakeLists.txt
```diff
@@ -9,28 +78,22 @@ if(CONFIG_BOOTLOADER_MCUBOOT)
...
-
   get_property(app_binary_dir GLOBAL PROPERTY PROJECT_BINARY_DIR)
   set(merged_hex_file
     ${app_binary_dir}/mcuboot_primary_app.hex)
   set(merged_hex_file_depends
     mcuboot_primary_app_hex$<SEMICOLON>${PROJECT_BINARY_DIR}/mcuboot_primary_app.hex)
   set(sign_merged
     $<TARGET_EXISTS:partition_manager>)
-  set(to_sign_hex
+  set(app_to_sign_hex
```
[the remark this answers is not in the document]
hakonfam: fixed

zephyr/CMakeLists.txt
hakonfam: I see your point. But after making the change, I find it harder to see the relation between `${slot}` and `${slot}_image`. 

Earlier you could easily see if the variable being used was that of the full Sx slot, or just the image. I agree that the distinction is not clear to everyone, but that could be fixed using some docs.

zephyr/CMakeLists.txt
hakonfam: fixed

zephyr/CMakeLists.txt
hakonfam: Without multi image I don't see any benefit of this, as the user would have to manually point to what hex file should be signed etc.

Also, upstream doesn't have mcuboot as a west-module, hence this file does not exist upstream.

zephyr/CMakeLists.txt
oyvindronningstad: Ok, how about this?

    foreach(slot_only s0;s1)
        set(slot ${slot_only}_image)

hakonfam: I agree

oyvindronningstad: Ok, good :). My suggestion:

    foreach(parent_slot s0;s1)
        set(slot ${parent_slot}_image)

PR #52, opened 2019-09-04
[nrf temphack] ci: Skip GitLint compliance check if PR-title has [nrf mergeup]

thst-nordic:
This CI plan will skip Linting if '[nrf mergeup]' is in the title. 
Also cherry-picked CVE from upstream which seemed urgent.

PR #53, opened 2019-09-11
[DNM] [nrf mergeup] Mergeup post zephyr 2.0

mbolivar:
A mergeup. Build tested only. Zephyr and nrf PRs forthcoming.


nvlsianpu:
I added `[nrf mergeup]` to the PR title as want to check whether #52 will works.

nvlsianpu:
^^ doesn't work
@thst-nordic is that expected, as initial pr name was diferent?

PR #44, opened 2019-06-28
MCU boot for pca20035

joakimtoe:

PR #55, opened 2019-09-20
Add Thingy91 support

bjda:
Enables mcuboot to work on the new Nordic Thingy:91.

This PR is a superset of #44 , which should be closed.


bjda:
@mbolivar @nvlsianpu @pdunaj ping

boot/zephyr/CMakeLists.txt
```diff
@@ -54,6 +54,9 @@ macro(app_set_runner_args)
   endif()
 endmacro()
 
+# nRF Connect SDK application boilerplate, needed for out-of-tree Nordic boards.
```
nvlsianpu: @SebastianBoe Is this line OK.  Can this interfere with boilerplate bellow?

boot/zephyr/Kconfig
```diff
@@ -174,7 +174,7 @@ config BOOT_MAX_IMG_SECTORS
 
 config BOOT_ERASE_PROGRESSIVELY
 	bool "Erase flash progressively when receiving new firmware"
-	default y if SOC_NRF52840
+	default y if SOC_NRF52840 || SOC_NRF9160
```
nvlsianpu: What the reason for enabling this for all nRF9160?

boot/zephyr/CMakeLists.txt
SebastianBoe: This is OK (and best-practice for NCS application's).

boot/zephyr/Kconfig
bjda: This avoids a ~10 second delay at the beginning of the flashing process. Given that is saves this much time in a typical use case, I considered it a sensible default.

mbolivar:
These seem fine to me, but shouldn't the `[nrf toup] zephyr: add support for Thingy:91` patch be `noup`, not `toup`? That board definition is downstream only, so I don't think upstream mcuboot would want that patch.

@nvlsianpu, what do you think?

bjda:
@mbolivar I was wondering about that. I expect the board itself will be upstreamed quite soon (it is public after all). If we should wait until after the fact, say the word and i will change it to `[noup]`.

mbolivar:
Hi @bjda -- I think it would be cleaner to keep this noup until the decision is definitely clear we will upstream the board. If that decision has been made, never mind me :).

joakimtoe:
@mbolivar : Yes, the board is to be upstreamed.

mbolivar:
> @mbolivar : Yes, the board is to be upstreamed.

In that case this looks fine to me as a temporary measure until we have time to upstream it out of the nrf repo. Thanks!

PR #58, opened 2019-10-01
[nrf noup] cmake: add fw_metadata overlay if SECURE_BOOT

hakonfam:
Also leverage new function 'add_overlay_config'.
This function does exactly what was inlined before.


hakonfam:
Will fix by adding `FW_METADATA=y` to `prj.conf` directly instead.

PR #59, opened 2019-10-01
[nrf fromlist]zephyr/serial_recovery: fix serial device binding

nvlsianpu:
zephyr changes the way in which it provide name of
serial device used by serial recover.
see: zephyrproject-rtos/zephyr#18752

This path introduce build time configuration of device name
which is set to exact the same device which was set before.


nvlsianpu:
cc @andvib

andvib:
Re-tested locally, this PR fixes the compilation issue

PR #68, opened 2019-11-05
cmake: scripts: generate zip files for app and mcuboot updates

hakonfam:
When enabled, this feature will generate a zip file for each
upgradeable component. This is the file which is typically given
to the DFU backend. The ZIP file also contains a 'manifest.json'
file which describes the contents.


hakonfam:
> No sauce tag and looks like it might be nrf specific?

Fixed!

hakonfam:
ping @oyvindronningstad  @sigvartmh

mbolivar:
How does this interact with https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1508?

Is it possible to put this script in nrf and have it called from the applications which generate FOTA packages?

hakonfam:
> How does this interact with [NordicPlayground/fw-nrfconnect-nrf#1508](https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1508)?

AFAICS this PR does not introduce any "extra" interactions between this #1508.
 
> Is it possible to put this script in nrf and have it called from the applications which generate FOTA packages?

Do you mean the .py script? I guess that makes sense, as it is not specific for mcuboot.

mbolivar:
> > Is it possible to put this script in nrf and have it called from the applications which generate FOTA packages?
> 
> Do you mean the .py script? I guess that makes sense, as it is not specific for mcuboot.

That would be ideal if possible then. We shouldn't be putting things into our downstream forks without good reason, as it creates a bigger maintenance / synchronization burden.

hakonfam:
Moved the script file to `nrf/scripts/bootloader/generate_zip.py`.

mbolivar:
>> Is it possible to put this script in nrf and have it called from the applications which generate FOTA packages?
> 
> Moved the script file to `nrf/scripts/bootloader/generate_zip.py`.

OK, thanks! So that's the first half ("put this script in nrf"), but what about "have it called from the applications which generate FOTA packages"? E.g. from `nrf/applications/asset_tracker/CMakeLists.txt`. Then no OOT patches are needed in mcuboot.

I'm just generally trying to understand why we need to keep adding downstream specific stuff into our mcuboot fork. Would doing what I'm asking about require waiting for @tejlmand 's new multi-image patch series? I was under the impression that that work would let you specify build outputs from one child image in your main application CMakeLists.txt, where they could be post-processed (e.g. to create the FOTA package).

hakonfam:
> OK, thanks! So that's the first half ("put this script in nrf"), but what about "have it called from the applications which generate FOTA packages"? E.g. from `nrf/applications/asset_tracker/CMakeLists.txt`. Then no OOT patches are needed in mcuboot.

AFAIK this entire file is a OOT patch? In other words its just a bunch of commits to drag along, no conflicts.

> I'm just generally trying to understand why we need to keep adding downstream specific stuff into our mcuboot fork.

I guess it is because this is the place we put all cmake code which needs to be executed when mcuboot is included in the build. One alternative could be to make a wrapper script in the `nrf` repo, and have all of this code there. I would love that change :)

Since the information needed to generate the zip file in this cmake-context, we can not move only the generate-zip part of the logic. Either we move the entire file, or we keep the generate-zip logic here.

hakonfam:
@mbolivar , I've made a proposal for moving this whole file here https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1576

mbolivar:
This repository's history has been rewritten for the v1.2 release. Please rebase this PR if still relevant, or close it if no longer needed. Thanks!

PR #70, opened 2019-11-14
[DNM] [nrf noup] Remove direct DT_ constants evaluation in CMakeList

de-nordic:
It is no longer possible to directly evaluete DT_ within CMakeList,
instead intermediate variable defined within Kconfig needs to be
used.

Signed-off-by: Dominik Ermel <dominik.ermel@nordicsemi.no>

### DIFF
--- a/zephyr/Kconfig
+++ b/zephyr/Kconfig
@@ -33,4 +33,7 @@ config BOOT_SIGNATURE_KEY_FILE
 	  with the public key information will be written in a format expected by
 	  MCUboot.
 
+config DT_FLASH_WRITE_BLOCK_SIZE
+	def_int $(dt_int_val,DT_FLASH_WRITE_BLOCK_SIZE)
+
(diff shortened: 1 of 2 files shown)


zephyr/Kconfig
```diff
@@ -33,4 +33,7 @@ config BOOT_SIGNATURE_KEY_FILE
 	  with the public key information will be written in a format expected by
 	  MCUboot.
 
+config FLASH_BLOCK_ALIGN
+	def_int $(dt_int_val,DT_FLASH_WRITE_BLOCK_SIZE)
```
SebastianBoe: Could we give it a name that is as similar as possible to the source? Then it would be easier for users to understand that this is the same thing.

e.g.

CONFIG_DT_FLASH_WRITE_BLOCK_SIZE

de-nordic: Done

PR #72, opened 2019-11-25
[nrf noup] boot: zephyr: Add check of RESETREAS for Thingy:91

jhn-nordic:
Only execute serial recovery if button is pressed during power
up reset or pin reset. This is to prevent going into serial
recovery when waking up from power down with button press.

Signed-off-by: Jon Helge Nistad <jon.helge.nistad@nordicsemi.no>

### DIFF
--- a/boot/zephyr/main.c
+++ b/boot/zephyr/main.c
@@ -44,6 +44,26 @@ const struct boot_uart_funcs boot_funcs = {
 #include <usb/class/usb_dfu.h>
 #endif
 
+#ifdef CONFIG_SOC_FAMILY_NRF
+#include <hal/nrf_power.h>
+
+static inline bool boot_skip_serial_recovery()
+{
+#if NRF_POWER_HAS_RESETREAS	
+    u32_t rr = nrf_power_resetreas_get(NRF_POWER);
+
+    return !(rr == 0 || (rr & NRF_POWER_RESETREAS_RESETPIN_MASK));
+#else
+    return false;
+#endif
+}
+#else
+static inline bool boot_skip_serial_recovery()
+{
+    return false;
+}
+#endif
+
 MCUBOOT_LOG_MODULE_REGISTER(mcuboot);
 
 void os_heap_init(void);
@@ -193,8 +213,8 @@ void main(void)
     rc = gpio_pin_read(detect_port, CONFIG_BOOT_SERIAL_DETECT_PIN,
                        &detect_value);
     __ASSERT(rc == 0, "Error of the reading the detect pin.\n");
-
-    if (detect_value == CONFIG_BOOT_SERIAL_DETECT_PIN_VAL) {
+    if (detect_value == CONFIG_BOOT_SERIAL_DETECT_PIN_VAL &&
+        !boot_skip_serial_recovery()) {
         BOOT_LOG_INF("Enter the serial recovery mode");
         rc = boot_console_init();
         __ASSERT(rc == 0, "Error initializing boot console.\n");


boot/zephyr/main.c
```diff
@@ -44,6 +44,10 @@ const struct boot_uart_funcs boot_funcs = {
 #include <usb/class/usb_dfu.h>
 #endif
 
+#ifdef CONFIG_BOARD_NRF9160_PCA20035
```
mbolivar: This looks more like `nrf temphack` than `nrf noup` to me.

Do the existing application specific hooks in `<power/power.h>` not allow this to be done in a more generic way? I can't see why we want this to be just for one particular upstream board, since any nRF-derived board (including customer designs) could want this same feature, no?

jhn-nordic: I do not find any thing in power.h for this use case, but it can most likely be done more generic for the the nRF devices. Something like below would probably be safe and generic?

#ifdef CONFIG_SOC_FAMILY_NRF
#if NRF_POWER_HAS_RESETREAS
....
#endif 
#endif

mbolivar: That's better! How about going all the way with something like this?

```c
#ifdef CONFIG_SOC_FAMILY_NRF
#include <hal/nrf_power.h>

static inline bool boot_skip_serial_recovery()
{
	u32_t rr = nrf_power_resetreas_get();

	return !(rr == 0 || (rr & NRF_POWER_RESETREAS_RESETPIN_MASK));
}
#else
static inline bool boot_skip_serial_recovery()
{
	return false;
}
#endif

/* ... */

if (detect_value == CONFIG_BOOT_SERIAL_DETECT_PIN_VAL && \
	!boot_skip_serial_recovery()) {

}
```

I bet that would be relatively easy to upstream into MCUboot, since 1. zephyr doesn't have a generic API for this you could find and 2. the vendor-specific portions are hidden away a bit more from the application code

jhn-nordic: ok. I did not see the comment before I updated the PR with a variant based on my first commit. I will try out what you propose when back in office tomorrow.

jhn-nordic:
Added DNM until I have actually tested the latest change on real device.

boot/zephyr/main.c
```diff
@@ -193,8 +213,8 @@ void main(void)
     rc = gpio_pin_read(detect_port, CONFIG_BOOT_SERIAL_DETECT_PIN,
                        &detect_value);
     __ASSERT(rc == 0, "Error of the reading the detect pin.\n");
-
-    if (detect_value == CONFIG_BOOT_SERIAL_DETECT_PIN_VAL) {
+    if (detect_value == CONFIG_BOOT_SERIAL_DETECT_PIN_VAL && \
```
mbolivar: Sorry, I know that's copy/pasted from my code, but I had python on the brain... backslash at end of line is not needed :)

boot/zephyr/main.c
mbolivar: Sounds good, thanks! I think you might as well submit it upstream if the approach looks OK to you so we don't have to carry the downstream patch.

jhn-nordic:
Made upstream PR
https://github.com/JuulLabs-OSS/mcuboot/pull/593

mbolivar:
> Made upstream PR
> [JuulLabs-OSS/mcuboot#593](https://github.com/JuulLabs-OSS/mcuboot/pull/593)

I've approved that one. As discussed offline, if you need this merged into NCS master urgently, and upstream review or the upmerge process is being too slow, we can do this as a fromlist or fromtree in NCS.

Thanks!

PR #73, opened 2019-11-26
cmake: move all logic from zephyr/CMakeLists.txt to nrf

hakonfam:
Do this to simplify working with multiple repositories.

See https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1576


SebastianBoe:
missing a tag

hakonfam:
ping @mbolivar @SebastianBoe @nvlsianpu

zephyr/CMakeLists.txt
```diff
@@ -1,187 +1 @@
...
-
-      # Override the container hex file target.
-      set_property(GLOBAL PROPERTY
-        ${parent_slot}_PM_TARGET
-        signed_${parent_slot}_target
-        )
-    endforeach()
-  endif()
-endif()
+include(../../nrf/cmake/mcuboot.cmake)
```
mbolivar: Sorry for the slow response -- last week there were 2 US holidays. Thanks a lot for working on this.

Can you explain why this needs to remain here? I naively assumed that invoking this logic could be done in the app/CMakeLists.txt for the root image, since I think that's the one that ultimately decides the boot chain and therefore what needs signing.

Not trying to get in the way, just trying to understand better.

Thanks a lot!

hakonfam: No problem :) 

AFAIK this is the piece of code which is executed when this repo is included as a module. 
It made sense to be to have this here as it seems that this file is executed automatically.

mbolivar: > No problem :)

:)
 
> AFAIK this is the piece of code which is executed when this repo is included as a module.

Yup, that's right.

> It made sense to be to have this here as it seems that this file is executed automatically.

I'm still confused about this issue from my first comment, sorry:

> I naively assumed that invoking this logic could be done in the app/CMakeLists.txt for the root image, since I think that's the one that ultimately decides the boot chain and therefore what needs signing.

hakonfam: > I'm still confused about this issue from my first comment, sorry:
> 
> > I naively assumed that invoking this logic could be done in the app/CMakeLists.txt for the root image, since I think that's the one that ultimately decides the boot chain and therefore what needs signing.

In that case, each application must include some boilerplate *if* it is going to be used with MCUBoot. Even if its only a single line of `include` I much prefer the solution proposed in this PR.

mbolivar: > In that case, each application must include some boilerplate _if_ it is going to be used with MCUBoot. Even if its only a single line of `include` I much prefer the solution proposed in this PR.

Summary of what we discussed offline:

1. we'll go with this for now
2. we can try moving this content into the module CMakeLists.txt for the nrf repository, instead of its current location (module CMakeLists.txt for the mcuboot repository, which may not be needed at all as a file after the new multi-image implementation goes in)

Did I get that right?

hakonfam: Yes!

tejlmand:
```suggestion
include(${ZEPHYR_NRF_MODULE_DIR})
```
cause I believe this is being executed as part of Zephyr module inclusion

tejlmand: On second thought.
Any special reason this include needs to happen here ?
Could this `include` code be moved here: https://github.com/NordicPlayground/fw-nrfconnect-nrf/blob/master/CMakeLists.txt#L10-L19
instead.

mbolivar:
This repository's history has been rewritten for the v1.2 release. Please rebase this PR if still relevant, or close it if no longer needed. Thanks!

hakonfam:
@tejlmand  your comment should be addressed

PR #75, opened 2020-01-15
[nrf noup] ci: adding common cron configuration

thst-nordic:
- Add west.yml in order to get the zephyr-rtos/ci-tools. 
- update .gitlint for new ci-tools path. 
- move to CI_LIB-v2 jenkinsfile format

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -10,7 +10,7 @@ debug = false
 
 # Set the extra-path where gitlint will search for user defined rules
 # See http://jorisroovers.github.io/gitlint/user_defined_rules for details
-extra-path=../ci-tools/scripts/gitlint
+extra-path=../tools/ci-tools/scripts/gitlint
 
 [title-max-length-no-revert]
 line-length=72
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,11 +1,24 @@
+//
+// Copyright (c) 2020 Nordic Semiconductor ASA. All Rights Reserved.
+//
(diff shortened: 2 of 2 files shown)


thst-nordic:
the Jenkinsfile changes are the same as https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1711 and https://github.com/NordicPlayground/fw-nrfconnect-zephyr/pull/257

west.yml
```diff
@@ -0,0 +1,37 @@
+# The west manifest file for the nRF Connect SDK (NCS).
```
mbolivar: This shouldn't be here.

thst-nordic: we were using CI_LIB to get/control the zephyr-rtos/ci-tools repo before.  I think using west.yml is a better.

carlescufi: I also think this is very misleading because you are creating a manifest file that only references ci-tools but not zephyr itself, while MCUboot uses Zephyr to build. If all you need is to hardcode a particular version of ci-tools then you can do that elsewhere, but I agree with @mbolivar that this should not be here, because our users will see a `west.yml` in the root of our MCUboot fork and they will assume that this is a standalone package that can be passed to `west init` in order to retrieve MUCboot and its dependencies (including zephyr) and to build them. This won't be the case, so it's abusing west.

thst-nordic: ok, I will remove and get ci-tools a different way.

PR #79, opened 2020-01-24
Remove scratch

hakonfam:
Get rid of not-required scratch area and use move-swap instead.

### DIFF
--- a/boot/zephyr/pm.yml
+++ b/boot/zephyr/pm.yml
@@ -20,11 +20,13 @@ mcuboot_secondary:
     align: {start: DT_FLASH_ERASE_BLOCK_SIZE}
     after: mcuboot_primary
 
+#ifndef CONFIG_BOOT_SWAP_USING_MOVE
 mcuboot_scratch:
   size: CONFIG_PM_PARTITION_SIZE_MCUBOOT_SCRATCH
   placement:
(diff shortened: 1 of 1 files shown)


boot/zephyr/prj.conf
```diff
@@ -45,3 +45,6 @@ CONFIG_LOG=y
 CONFIG_LOG_IMMEDIATE=y
 ### Ensure Zephyr logging changes don't use more resources
 CONFIG_LOG_DEFAULT_LEVEL=0
+
+# Save space by not using scratch partition, but smart swap instead.
+CONFIG_BOOT_SWAP_USING_MOVE=y
```
nvlsianpu: y will be the default value for NRF, `temphack` or remove?

hakonfam:
Closing as this will be fixed upstream

hakonfam:
Forgot what this PR was about, will remove the commit now in ustream.

hakonfam:
IMO there is no need to wait for upmerge, as this has the desired effect in the existing code. The user can enable this option already, its just not the default.

nvlsianpu:
OK. so lets wait for current upmerge

nvlsianpu:
@hakonfam bump version in nrf manifest

PR #83, opened 2020-02-07
[nrf fromlist] kconfig: add default pin for serial recovery for nrf53

hakonfam:
Prior to this the kconfig would fail because of invalid value for an int

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -305,6 +305,7 @@ config BOOT_SERIAL_DETECT_PIN
 	default 6 if BOARD_NRF9160_PCA10090
 	default 11 if BOARD_NRF52840_PCA10056
 	default 13 if BOARD_NRF52_PCA10040
+	default 23 if BOARD_NRF5340_DK_NRF5340_CPUAPP || BOARD_NRF5340_DK_NRF5340_CPUAPPNS
 	help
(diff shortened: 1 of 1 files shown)


hakonfam:
I guess this should be upstreamed, but left here for initial review.

boot/zephyr/Kconfig
```diff
@@ -305,6 +305,7 @@ config BOOT_SERIAL_DETECT_PIN
 	default 6 if BOARD_NRF9160_PCA10090
 	default 11 if BOARD_NRF52840_PCA10056
 	default 13 if BOARD_NRF52_PCA10040
+	default 23 if BOARD_NRF5340_DK_NRF5340_CPUAPP || BOARD_NRF5340_DK_NRF5340_CPUAPPNS
```
nvlsianpu: strange that same pin is used for both

hakonfam: It is only the non-secure alias for the same SoC, hence the same pin (it is button 1 on the 53 DK)

hakonfam:
@nvlsianpu changed to fromlist

boot/zephyr/Kconfig
nvlsianpu: good, thanks for clarification

hakonfam:
This is in upstream now, and will be included in next upmerge.

PR #85, opened 2020-02-17
loader: Fixed warning about uninitialized rc variable

erik-johnson-nimbelink:
Might get past the loop without going in at least once.

Signed-off-by: Erik Johnson <erik.johnson@nimbelink.com>

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1189,7 +1189,7 @@ done:
 static int
 boot_verify_dependencies(struct boot_loader_state *state)
 {
-    int rc;
+    int rc = -1;
     uint8_t slot;
 
     BOOT_CURR_IMG(state) = 0;


hakonfam:
Thank you for the contribution! My first thought is that we should send this patch upstream, as this seems like an issue in the upstream repository as well. 

Nordic maintains a fork of mcuboot, but we strive to keep the diff as small as possible.
So changes like these, which are not specific to nordic, should ideally go directly upstream.

Once the upstream PR is up, we can create a [nrf fromlist] commit in this fork, paired with a manifest update PR in the fw-nrfconnect-nrf repo (I know, its complicated :D ) to get the change in faster than the regular upmerge.

Let me know if you feel that this is too much work and you want me to do it for you, multi repo can be a pain :) It also looks like we are short on public documentation on how to actually contribute to our fork.

erik-johnson-nimbelink:
No worries, I can do the work on this. I've already contributed to the Zephyr project on behalf of something I found while using the nRF Connect SDK, so I'm familiar with the workflow. I just didn't know (or forgot) the MCUBoot repo was a minimal fork of an upstream project.

Should I report back to this PR with the one for upstream? Or will you just pick it up when it gets merged in?

Thanks!

hakonfam:
Once the upstream PR is opened it is a bit up to you. Our fork is not upmerged that frequently, so what we do depends on how fast you want the change (to me it does not look like a blocker).

If you are not in a hurry, just create the PR upstream, and then the change will trickle down eventually as we do regular upmerges for our fork.

If you want the change to get in ASAP just keep this PR and change the commit message into 
[nrf fromlist] and then tag the PR number in the commit message.

erik-johnson-nimbelink:
Definitely not a show-stopper for us, just figured I'd quiet the warning. Happy to close this one, then.

PR #96, opened 2020-05-13
[nrf fromtree] serial_recovery: Reduce minimum number of members in packet

oyvindronningstad:
Was 3, but mcumgr sometimes sends 2 ("off" + "data", "len" is not needed
when "off" != 0). Reduce to 1 to avoid the problem, since the code
has other checks to catch this.

Ref: DESK-939

Signed-off-by: Øyvind Rønningstad <oyvind.ronningstad@nordicsemi.no>

### DIFF
--- a/boot/boot_serial/src/serial_recovery.cddl
+++ b/boot/boot_serial/src/serial_recovery.cddl
@@ -11,5 +11,5 @@ Member = ("image" => int) /
 	("sha" => bstr)
 
 Upload = {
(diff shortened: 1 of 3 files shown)


oyvindronningstad:
@pdunaj can you try it out?

pdunaj:
I have cherry-picked this change and tested with nrf_desktop serial recovery. Works fine :)

PR #103, opened 2020-05-27
[nrf temphack] Ensure that MCUBoot read buffers are word aligned

sigvartmh:
Since other flash drivers does not handle unaligned access yet, we
have to ensure that the read buffers are word aligned. We'll at most
throw away 6 bytes so I think it should be okei.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -69,7 +69,7 @@ static struct boot_loader_state boot_data;
  * to just make those variables stack allocated.
  */
 #if !defined(__BOOTSIM__)
-#define TARGET_STATIC static
+#define TARGET_STATIC __aligned(4) static
 #else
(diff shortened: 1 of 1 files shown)


pabigot:
Bizarre.  My review did not include the comment I made at the code location, which was to the effect that there are upwards of 50 locations in mcuboot where a buffer is passed to `flash_area_{read,write}` that should be audited to ensure the buffers passed have 4-byte alignment.  Several do not.

The comment also said it was reasonable to leave it with this solution for now, and plan to either do a full assessment of mcuboot and mcumgr after the release, or complete the task of making sure Zephyr supports byte alignment for all flash API buffers, offsets, and sizes.

hakonfam:
DNM while waiting for upmerge

tejlmand:
Included in #104

PR #104, opened 2020-05-28
[nrf fromlist] cmake: use find_package to locate Zephyr

tejlmand:
[nrf fromlist] cmake: use find_package to locate Zephyr

Using find_package to locate Zephyr.

Old behavior was to use $ENV{ZEPHYR_BASE} for inclusion of boiler plate
code.

Whenever an automatic run of CMake happend by the build system / IDE
then it was required that ZEPHYR_BASE was defined.
Using ZEPHYR_BASE only to locate the Zephyr package allows CMake to
cache the base variable and thus allowing subsequent invocation even
if ZEPHYR_BASE is not set in the environment.

It also removes the risk of strange build results if a user switchs
between different Zephyr based project folders and forgetting to reset
ZEPHYR_BASE before running ninja / make.

Signed-off-by: Torsten Rasmussen <Torsten.Rasmussen@nordicsemi.no>

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -69,7 +69,7 @@ static struct boot_loader_state boot_data;
  * to just make those variables stack allocated.
  */
 #if !defined(__BOOTSIM__)
-#define TARGET_STATIC static
+#define TARGET_STATIC __aligned(4) static
 #else
 #define TARGET_STATIC
 #endif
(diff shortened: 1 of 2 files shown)


tejlmand:
Upstream PRs:
https://github.com/zephyrproject-rtos/mcuboot/pull/22
https://github.com/JuulLabs-OSS/mcuboot/pull/745

PR #108, opened 2020-07-13
boot: zephyr: Use private key

MarekPieta:
Change adds using own private key for MCUBoot bootloader.


hakonfam:
Is it not possible to set `CONFIG_BOOT_SIGNATURE_KEY_FILE ` somewhere in your own cmake code instead of adding these downstream changes?

MarekPieta:
Hi @hakonfam 
Applications need the possibility to use own private keys that are located in the application source directory.

Using absolute path is really not convenient for multiple collaborators (path depends on given PC/laptop).
Creating relative path from MCUBoot directory is not clean solution too - someone may e.g. copy the application sources to make his own changes without changing the key pathes. In that case application will still use old keys and it would be difficult to notice the problem (since we use relative path from MCUBoot directory).

I also make some followup changes to nrf cmake files: https://github.com/nrfconnect/sdk-nrf/pull/2577. For example I added warning to make sure that user will be aware that debug keys are in use.

Also I think that the change is not nrf specific. It uses CONF_FILE path. It could be useful for other Zephyr-based applications with MCUBoot.

nvlsianpu:
@MarekPieta 
This modification looks sane to me. Would be great to have note in kconfig help on file path type resolving precedence.
Can you open mirror PR to the upstream (once path accepted here)?

nvlsianpu:
@hakonfam Can you response to Marek comments?

hakonfam:
It is not clear to me what we gain from this feature. Sorry for being slow :)

> Applications need the possibility to use own private keys that are located in the application source directory.

Why? AFAICS you only need use a non-debug key when going into production, and in that case, you would not store the key in the source directory.

> Using absolute path is really not convenient for multiple collaborators (path depends on given PC/laptop).

I don't see why you need separate keys while the code is being developed, as these keys are checked in the repository alongside the code, and using different keys gives no increased security.

> Also I think that the change is not nrf specific. It uses CONF_FILE path. It could be useful for other Zephyr-based applications with MCUBoot.

The `CONF_FILE` variable should then be a cmake cache variable which has some documentation attached to it IMO.
I'm not sure about the current work flow, but typically the upstream PR needs to be posted before the downstream one goes in, this to make the diff as small as possible.

MarekPieta:
@hakonfam, I think that using private keys during development and storing them in source directory may depend on usecase.
For example the application may be developed only on internal branch/fork (so only contributors have access the files - then we may store the key next to source files). Also the commit only adds new functionality, there is no drawback related to adding this change.
I will add information to Kconfig help (following @nvlsianpu suggestions), so everything should be clear.

zephyr/Kconfig
```diff
@@ -39,6 +39,11 @@ config BOOT_SIGNATURE_KEY_FILE
 	string "PEM key file"
 	default "root-rsa-2048.pem"
 	help
+	  You can use either absolute or relative path.
+	  In case relative path is used, the build system assumes that it starts
+	  from the directory where the MCUBoot configuration file is located.
```
hakonfam, line 44: You should probably explain a bit more about the `MCUBoot configuration file` it might not be clear that it is the selected Kconfig configuration file.

boot/zephyr/CMakeLists.txt
```diff
@@ -210,9 +210,13 @@ if(CONFIG_MCUBOOT_SERIAL)
     )
 endif()
 
+get_filename_component(CONF_DIR ${CONF_FILE} DIRECTORY)
```
hakonfam, line 213: Since `CONF_FILE` is not used elsewhere in MCUBoot repository, you might want to add a comment stating that `CONF_FILE` points to the kconfig configuration file (e.g. prj.conf)

boot/zephyr/CMakeLists.txt
```diff
@@ -210,9 +210,13 @@ if(CONFIG_MCUBOOT_SERIAL)
...
+
 if(NOT CONFIG_BOOT_SIGNATURE_KEY_FILE STREQUAL "")
   if(IS_ABSOLUTE ${CONFIG_BOOT_SIGNATURE_KEY_FILE})
     set(KEY_FILE ${CONFIG_BOOT_SIGNATURE_KEY_FILE})
+  elseif(EXISTS ${CONF_DIR}/${CONFIG_BOOT_SIGNATURE_KEY_FILE})
```
hakonfam, line 218: To be absolutely sure that you are not interfering with existing functionality, but instead extending, you should consider moving this check below the check from the base of the repository.

zephyr/Kconfig
MarekPieta: Ok, I will add information that this relates to KConfig configuration file.

boot/zephyr/CMakeLists.txt
MarekPieta: To be honest I prefer to check application configuration directory first. I think that dedicated application keys should have precedence (since they are dedicated for the given application). The default keys should be used only if no other keys are specified.

boot/zephyr/CMakeLists.txt
MarekPieta: ok

boot/zephyr/CMakeLists.txt
hakonfam: Fine with me :)

hakonfam:
Can you create an upstream PR so you can add a reference and sauce tag? (i.e. [nrf fromlist])

MarekPieta:
Upstream PR:
https://github.com/JuulLabs-OSS/mcuboot/pull/781

nvlsianpu:
@hakonfam We have this merged in upsteram now.

MarekPieta:
@nvlsianpu, @hakonfam Can we merge this one ?

PR #112, opened 2020-08-03
Support for nrf52840 with ecc keys and cryptocell

piociu:


ext/nrf/cc310_glue.h
```diff
@@ -13,6 +13,13 @@
 #include <devicetree.h>
 #include <string.h>
 
+/*
+ * Name translation for peripherals with only one type of access available.
+ */
+#ifndef NRF_TRUSTZONE_NONSECURE
```
sigvartmh, line 19: Does this work for `NRF52` because I don't think it will(I could be wrong).
Since `NRF_TRUSTZONE_NONSECURE` is not defiend but NRF_CRYPTOCELL_S is not a register name in `NRF52`

piociu: This translation only works if NRF_TRUTZONE_NONSECURE is defined, otherwise (nrf52) translation is not needed. See also changes in lines 41 and 46.

sigvartmh:
```suggestion
#ifdef NRF_TRUSTZONE_NONSECURE

piociu: I get it now, condition updated with CONFIG_ARM_TRUSTZONE_M check

piociu:
I opened a PR in upstream: https://github.com/JuulLabs-OSS/mcuboot/pull/780

mbolivar-nordic:
> I opened a PR in upstream: [JuulLabs-OSS/mcuboot#780](https://github.com/JuulLabs-OSS/mcuboot/pull/780)

Thanks. Looks like that is merged now. @nvlsianpu should this be merged now or do you think we should wait for the next upmerge?

piociu:
I need this patch very soon. I would prefer to have it merged now.

nvlsianpu:
@piociu Can you open manifest PR for trigger NRF CI?

piociu:
PR in sdk-nrf: https://github.com/nrfconnect/sdk-nrf/pull/2660/

PR #113, opened 2020-08-04
boot: zephyr: Use correct pm template

pdunaj:
When single image DFU is used scratch is not needed.

Signed-off-by: Pawel Dunaj <pawel.dunaj@nordicsemi.no>

### DIFF
--- a/boot/zephyr/pm.yml
+++ b/boot/zephyr/pm.yml
@@ -20,7 +20,7 @@ mcuboot_secondary:
     align: {start: DT_FLASH_ERASE_BLOCK_SIZE}
     after: mcuboot_primary
 
-#ifndef CONFIG_BOOT_SWAP_USING_MOVE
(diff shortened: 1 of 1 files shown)


boot/zephyr/pm.yml
```diff
@@ -21,12 +21,14 @@ mcuboot_secondary:
     after: mcuboot_primary
 
 #ifndef CONFIG_BOOT_SWAP_USING_MOVE
+#ifndef CONFIG_SINGLE_IMAGE_DFU
```
hakonfam, line 24:
```suggestion
#if !defined(CONFIG_BOOT_SWAP_USING_MOVE) && !defined(CONFIG_SINGLE_IMAGE_DFU)
```

Do you think that this is more readable? Just a thought.

pdunaj: I don't have preference in this case. @nvlsianpu ?

nvlsianpu: what Håkon had suggested looks more legible to me.

pdunaj: changed

PR #115, opened 2020-08-12
[nrf toup] boot: zephyr: Fix build issue for multiple conf files

MarekPieta:
Change fixes build issue that occurs if CONF_FILE is a list of file paths instead of single file path.


nvlsianpu:
manifest PR https://github.com/nrfconnect/sdk-nrf/pull/2642/files

PR #116, opened 2020-08-17
boot: zephyr: nrf53 network core bootloader implementation

sigvartmh:
Enables network core updates of nrf53 using MCUBoot by identifying
images through their start addresses. Also implements the control and
transfer using the PCD module.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -712,10 +712,9 @@ boot_validated_swap_type(struct boot_loader_state *state,
     swap_type = boot_swap_type_multi(BOOT_CURR_IMG(state));
     if (BOOT_IS_UPGRADE(swap_type)) {
         /* Boot loader wants to switch to the secondary slot.
-         * Ensure image is valid. This checks header, hash/signature and trailer
+         * Ensure image is valid.
          */
         rc = boot_validate_slot(state, BOOT_SECONDARY_SLOT, bs);
-	/* Check for network update type */
         if (rc == 1) {
             swap_type = BOOT_SWAP_TYPE_NONE;
         } else if (rc != 0) {
@@ -728,8 +727,8 @@ boot_validated_swap_type(struct boot_loader_state *state,
 	if (reset_addr > PM_nrf5340pdk_nrf5340_cpunet_B0N_ADDRESS) {
 		uint32_t fw_size = hdr->ih_img_size;
 
-		BOOT_LOG_INF("Starting netowrk core update");
-		rc = do_network_core_update(NULL, vtable, fw_size);
+		BOOT_LOG_INF("Starting network core update");
+		rc = do_network_core_update(vtable, fw_size);
 		if (rc != 0) {
 			swap_type = BOOT_SWAP_TYPE_FAIL;
 		} else {
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -724,7 +724,7 @@ boot_validated_swap_type(struct boot_loader_state *state,
 	vtable_addr = (uint32_t)hdr + hdr->ih_hdr_size;
 	vtable = (uint32_t *)(vtable_addr);
 	reset_addr = vtable[1];
-	if (reset_addr > PM_nrf5340pdk_nrf5340_cpunet_B0N_ADDRESS) {
+	if (reset_addr > PM_CPUNET_B0N_ADDRESS) {
 		uint32_t fw_size = hdr->ih_img_size;
 
 		BOOT_LOG_INF("Starting network core update");
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -45,6 +45,10 @@
 #include "bootutil/security_cnt.h"
 #include "bootutil/boot_record.h"
 
+#ifdef CONFIG_SOC_SERIES_NRF53X
+#include "nrf53_cpunet_ctl.h"
+#endif
+
 #ifdef MCUBOOT_ENC_IMAGES
 #include "bootutil/enc_key.h"
 #endif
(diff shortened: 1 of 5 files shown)


sigvartmh:
Not 100% complete I have a few questions on what to do.

boot/zephyr/main.c
```diff
@@ -326,6 +330,9 @@ void main(void)
     BOOT_LOG_INF("Starting bootloader");
 
     os_heap_init();
+#if defined(CONFIG_SOC_SERIES_NRF53X) && defined(CONFIG_DEBUG)
+    enable_network_core_debug_pins();
```
sigvartmh, line 334: We could just leave `CONFIG_ENABLE_NETCORE` enabled and it will include similar functionality. However we would then start MCUBoot with a state where the network core is already online.

This means we would have to change `boot/zephyr/nrf53_cpunet_ctl.c:21 to first turn it off then turn it on.

boot/zephyr/nrf53_cpunet_ctl.c
```diff
@@ -0,0 +1,81 @@
...
+	GPIO_PIN_CNF_MCUSEL_NetworkMCU << GPIO_PIN_CNF_MCUSEL_Pos;
+
+	/* Route Bluetooth Controller Debug Pins */
+	DEBUG_SETUP();
+
+	/* Retain nRF5340 Network MCU in Secure domain (bus
+	 * accesses by Network MCU will have Secure attribute set).
+	 */
+	NRF_SPU->EXTDOMAIN[0].PERM = 1 << 4;
+}
```
sigvartmh, line 81: If we leave the network core on at boot we can get away with not having this code for enabling uarte output pins for the network core as it is done in the initialization code. See https://github.com/nrfconnect/sdk-zephyr/blob/master/boards/arm/nrf5340dk_nrf5340/nrf5340_cpunet_reset.c

boot/bootutil/src/loader.c
```diff
@@ -703,14 +712,32 @@ boot_validated_swap_type(struct boot_loader_state *state,
...
+		uint32_t fw_size = hdr->ih_img_size;
+
+		BOOT_LOG_INF("Starting netowrk core update");
+		rc = do_network_core_update(NULL, vtable, fw_size);
+		if (rc != 0) {
+			swap_type = BOOT_SWAP_TYPE_FAIL;
+		} else {
+			BOOT_LOG_INF("Done updateing network core");
+			BOOT_SWAP_TYPE(state) = swap_type = BOOT_SWAP_TYPE_NONE;
+			rc = swap_erase_trailer_sectors(state, secondary_fa);
```
sigvartmh, line 738: Should we keep the trailer and just change the flag from TEST to NONE?

oyvindronningstad:
Did we plan to lock the PCD_CMD_ADDRESS area at some point?

sigvartmh:
You mean flash lock it?

oyvindronningstad:
Yes, disable writes with SPU.

sigvartmh:
I don't see how that is achievable, but I can try to think about it a bit more. If we lock the area network core will no longer be able to report back if we lock it after we have set it. So after we write `PCD_CMD_COPY` we wait for it to be `0` or `PCD_CMD_FAILURE` which are communicated through the same area.

We could probably set up an IPC signal to enable/disable write to the area however that will only secure it as much as it is currently as you could easily glitch it at the right moment. So I need to think a bit more about how we could do that if it is needed.

boot/zephyr/nrf53_cpunet_ctl.c
```diff
@@ -6,32 +12,47 @@
...
+	return false;
+}
+
+static bool successful(void)
+{
+	struct pcd_cmd *cmd = (struct pcd_cmd *)PCD_CMD_ADDRESS;
+	if (cmd->magic != PCD_CMD_MAGIC_DONE) {
+		return false;
+	}
+	return true;
```
oyvindronningstad, line 40: Nit: It's weird that the parity of this function is opposite of `is_copying`. It could also be shorter.
```suggestion
	return (cmd->magic == PCD_CMD_MAGIC_DONE);
```

sigvartmh: True I'll fix it up it was sort of meant to be temporary.

boot/zephyr/nrf53_cpunet_ctl.c
```diff
@@ -42,47 +63,8 @@ int do_network_core_update(void *addr, void *src_addr, size_t len)
...
-	/* Route Bluetooth Controller Debug Pins */
-	DEBUG_SETUP();
-
-	/* Retain nRF5340 Network MCU in Secure domain (bus
-	 * accesses by Network MCU will have Secure attribute set).
-	 */
-	NRF_SPU->EXTDOMAIN[0].PERM = 1 << 4;
+void lock_ipc_ram_with_spu(){
+	nrf_spu_ramregion_set(NRF_SPU,
+			      APP_CORE_SRAM_SIZE/CONFIG_NRF_SPU_RAM_REGION_SIZE,
```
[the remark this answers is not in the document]
sigvartmh: Good idea I'll see if it's possible.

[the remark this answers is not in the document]
sigvartmh: Should be done now

boot/zephyr/nrf53_cpunet_ctl.c
```diff
@@ -6,32 +12,47 @@
...
-	}
-
-	struct pcd_cmd *cmd = (struct pcd_cmd *)addr;
-
-	rsp->magic = PCD_CMD_MAGIC_COPY;
+	struct pcd_cmd *cmd = (struct pcd_cmd *)PCD_CMD_ADDRESS;
 	cmd->magic = PCD_CMD_MAGIC_COPY;
 	cmd->src_addr = src_addr;
 	cmd->len = len;
 	cmd->offset = 0x10800;
```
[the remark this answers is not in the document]
sigvartmh: fixed.

boot/bootutil/src/loader.c
sigvartmh: Just keeping it as is

PR #126, opened 2020-10-09
[nrf fromtree] fixed: swap-move, bootstrapping and fast overwrite mode

nvlsianpu:
* fix swap-move brick with padded image0

When the image in the primary slot is padded, the boot source is
considered the primary slot; this results in skipping the typical
initialization of the trailer, which ends up bricking the device. As
it is fine to always initialize the trailer in the primary slot when
starting a new upgrade the extra check was removed.

* fix boostrapping in swap-move

Fix boostrapping in swap-move that was being skipped due to the having
an erased header in the primary slot which caused an early return
because of the requirement of having to re-read image headers when
"moving" an image during an upgrade.

* copy image size with fast overwrite 

Previously when `MCUBOOT_OVERWRITE_ONLY_FAST` was set, the whole amount
of sectors that stored an image were being copied. After this commit
only the exact amount of data used by the image is copied; this avoids
copying some 0xff (or garbage) data between the end of the image and the
end of the last sector storing it.

Extra trailer management was added which suits using the copy upgrade
routine also for bootstrapping.

Signed-off-by: Fabio Utzig <fabio.utzig@nordicsemi.no>

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -961,6 +961,13 @@ boot_copy_image(struct boot_loader_state *state, struct boot_status *bs)
     const struct flash_area *fap_secondary_slot;
     uint8_t image_index;
 
+#if defined(MCUBOOT_OVERWRITE_ONLY_FAST)
+    uint32_t sector;
+    uint32_t trailer_sz;
+    uint32_t off;
+    uint32_t sz;
+#endif
+
     (void)bs;
 
 #if defined(MCUBOOT_OVERWRITE_ONLY_FAST)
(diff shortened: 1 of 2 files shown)


nvlsianpu:
@utzig Are https://github.com/JuulLabs-OSS/mcuboot/pull/826/ `botutill` part also relevant to be cherry-picked as bugfix here? - It is not clear for me if issues were imposed by FIH introduction or not.

utzig:
> @utzig Are [JuulLabs-OSS/mcuboot#826](https://github.com/JuulLabs-OSS/mcuboot/pull/826) `botutill` part also relevant to be cherry-picked as bugfix here? - It is not clear for me if issues were imposed by FIH introduction or not.

Yes, cherry-pick https://github.com/JuulLabs-OSS/mcuboot/pull/826/commits/738c8b7e47ea934218f5a9b1f880082a69001e97 and https://github.com/JuulLabs-OSS/mcuboot/pull/826/commits/b5801eea40ae005aaa5d1286038d6870539b3c6f.

utzig (approved):
LGTM

hakonfam:
Should the commits contain the SHA of the upstream commit? To make the upmerge process easier?

hakonfam (approved):
Since these are from upstream, and we trust our upstream (<3) I would say that they are OK. However, should we provide some TL;DR on who would be affected by this bricking bug being fixed? Either in release notes or somewhere else. Also check my comment regarding mentioning the commit SHA of the upstream commits in the downstream commits.

PR #119, opened 2020-09-09
zephyr/Kconfig: remove SINGLE_IMAGE_DFU duplication

nvlsianpu:
SINGLE_IMAGE_DFU option is already defined in boot/zephyr/Kconfig.

Kconfig: move BOOT_SIGNATURE_KEY_FILE to its upstream location

Signed-off-by: Andrzej Puzdrowski <andrzej.puzdrowski@nordicsemi.no>

PR #129, opened 2020-10-27
[nrf noup] Kconfig: create duplicate of sign key

hakonfam:
This duplicate is needed for the parent image to have access
to the signing key when MCUBoot is not built from source.

Ref: NCSDK-6859

### DIFF
--- a/zephyr/Kconfig
+++ b/zephyr/Kconfig
@@ -33,6 +33,25 @@ config MCUBOOT_FLASH_WRITE_BLOCK_SIZE
 	int
 	default $(dt_node_int_prop_int,$(DT_CHOSEN_ZEPHYR_FLASH),write-block-size)
 
+config BOOT_SIGNATURE_KEY_FILE
+	# Duplicate from boot/zephyr/Kconfig to expose this option to the parent
+	# image. Without this copy, it is not possible for the parent image to
+	# be signed if the MCUBoot child image is not built from source.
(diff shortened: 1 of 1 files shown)


hakonfam:
This is needed since I asked @nvlsianpu to remove the duplicate here: https://github.com/nrfconnect/sdk-mcuboot/pull/119#pullrequestreview-497344313

@mbolivar-nordic  I was thinking about leveraging the kconfig variable introduced to solve the same issue with `west sign` but my gut feeling was to revert this change, since its how it used to work before. Any thoughts?

mbolivar-nordic:
>  Any thoughts?

I'm really losing the bubble on what's going on with these; let's talk offline

tejlmand (changes requested):
I would prefer to revert all NCS specific patches, like: https://github.com/nrfconnect/sdk-mcuboot/commit/04514d6a2236050ac8abd89d72f59dd175a0265e regarding Kconfig changes, and move all the Kconfig logic into https://github.com/nrfconnect/sdk-nrf

Will be happy to assist on that work if you agree.

hakonfam:
Note to self, verify that the new solution sets the default value in the application.
This can be done building a project with MCUBOOT without setting the key explitcitly.

hakonfam:
Will be fixed using MODULE_ROOT functionality.

PR #136, opened 2021-01-19
DNM [nrf mergeup] Upmerge 180121

jfischer-no:
Upmerge 180121

Synchronize mcuboot up to upstream commit
3f49b5abf38659c22b6b73f5fe289de2e395263e

PR #137, opened 2021-01-20
[nrf noup] boot: zephyr: Separate minimal config from external crypto

kyberdin:
* Minimal image production has been moved into a prj_minimal.conf file to accommodate board-specific needs.
* Partition sizing has been moved into Kconfig to be consistent with the method used by b0.
* The minimal fragment has been simplified for only external crypto, and has been renamed to be consistent with other boot/zephyr fragments.
* Using this fragment with prj_minimal.conf makes MCUboot < 16kB for all nRF devices (9160 still needs 32kB partition).

Ref: NCSDK-6704
Signed-off-by: Stephen Stauts <stephen.stauts@nordicsemi.no>


nvlsianpu:
DNM before https://github.com/nrfconnect/sdk-mcuboot/pull/136

boot/zephyr/Kconfig
```diff
@@ -15,9 +15,12 @@ config MCUBOOT
 	select MPU_ALLOW_FLASH_WRITE if ARM_MPU
 	select USE_DT_CODE_PARTITION if HAS_FLASH_LOAD_OFFSET
 
-partition=MCUBOOT
-partition-size=0xc000
-source "${ZEPHYR_BASE}/../nrf/subsys/partition_manager/Kconfig.template.partition_size"
+config BOOT_MIN_PARTITION_SIZE
+    bool "Build for a minimal image size"
+
+#partition=MCUBOOT
```
hakonfam, line 21: I guess these should be removed?

boot/zephyr/Kconfig
```diff
@@ -27,6 +30,17 @@ partition=MCUBOOT_PAD
 partition-size=0x200
 source "${ZEPHYR_BASE}/../nrf/subsys/partition_manager/Kconfig.template.partition_size"
 
+config PM_PARTITION_SIZE_MCUBOOT
+    hex "Flash space reserved for the MCUboot image"
+    default 0xc000 if !BOOT_MIN_PARTITION_SIZE
+    default FPROTECT_BLOCK_SIZE if SOC_NRF9160
+    # nRF5340 can't shrink image into a single block without external crypto
```
hakonfam, line 37: What block? Maybe we should explain that the SPU is used for locking the flash, and that the smallest lockable unit is 16kb?

kyberdin:
@nvlsianpu I'm a bit confused about the CI errors:

1. Which CODEOWNERs file is it referring to?
2. I don't understand why the Kconfig options were not found. Did I break the Kconfig file?

Please advise.

nvlsianpu:
@thst-nordic Can you help with above? We don't use CODEOWNERS here so far. Also Kconfig failure looks false positive.

hakonfam:
I think we should make a PR in nrf which pulls this in. This to verify that it passes CI correctly.

hakonfam (approved):

kyberdin:
@hakonfam Opened related [sdk-nrf PR 3805](https://github.com/nrfconnect/sdk-nrf/pull/3805).

thst-nordic:
@nvlsianpu  the CODEOWNER test case is already being skipped for this repo. 
The Kconfig errors look to be the same variables being changed.  They look relevant, but I'm not certain.

carlescufi:
> @nvlsianpu the CODEOWNER test case is already being skipped for this repo.
> The Kconfig errors look to be the same variables being changed. They look relevant, but I'm not certain.
> 
> Maybe this PR is connected [nrfconnect/sdk-zephyr#438](https://github.com/nrfconnect/sdk-zephyr/pull/438)

No, that PR is for compliance in sdk-zephyr only. Maybe we should switch to using GitHub actions here too. I will try to take a look.

boot/zephyr/Kconfig
```diff
@@ -28,6 +27,19 @@ partition=MCUBOOT_PAD
 partition-size=0x200
 source "${ZEPHYR_BASE}/../nrf/subsys/partition_manager/Kconfig.template.partition_size"
 
+config PM_PARTITION_SIZE_MCUBOOT
+	hex "Flash space reserved for the MCUboot image"
+	default 0xc000 if !BOOT_MIN_PARTITION_SIZE
```
tejlmand, line 32: I'm not sure having a `BOOT_MIN_PARTITION_SIZE` setting is the best thing here.

This pattern will start to give n-kconfig symbols, one for each `minimal`, `external`, `mycustom`, `...` variant.

It's is already possible to do:
```
CONFIG_PM_PARTITION_SIZE_MCUBOOT=0x7c00
```
in the `prj_minimal.conf`

And then just keep `0xc000` as default if nothing is specified.

boot/zephyr/Kconfig
```diff
@@ -28,6 +27,19 @@ partition=MCUBOOT_PAD
...
+	default FPROTECT_BLOCK_SIZE if SOC_NRF9160
+	# nRF5340 uses SPU to lock flash, which uses lockable regions of 16kB.
+	# This image cannot be shrunk into a single region of this size without
+	# enabling external crypto.
+	default FPROTECT_BLOCK_SIZE if SOC_NRF5340_CPUAPP && BOOT_NRF_EXTERNAL_CRYPTO
```
tejlmand, line 37: The comment combined with this code sounds like the prompt should be disabled in this case, to avoid reconfiguration.
Remember, a default value is not locked, it's just a default if nothing else is configured.

tejlmand (changes requested):

nvlsianpu:
>@nvlsianpu the CODEOWNER test case is already being skipped for this repo.
The Kconfig errors look to be the same variables being changed. They look relevant, but I'm not certain.

>Maybe this PR is connected nrfconnect/sdk-zephyr#438

A few of these Kconfig properties are defined in sdk-zephyr, rest of them are there, so something is wrong with kconfig check.

boot/zephyr/Kconfig
```diff
@@ -28,6 +27,19 @@ partition=MCUBOOT_PAD
...
+	default 0x3e00 if BOOT_NRF_EXTERNAL_CRYPTO
```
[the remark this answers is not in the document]
kyberdin: [Line 140](https://github.com/nrfconnect/sdk-mcuboot/blob/cd7352c47c414276f956e90b0dd6021551cec76d/boot/zephyr/Kconfig#L140).

boot/zephyr/Kconfig
kyberdin: The idea here was to align the partition image sizing scheme here with [b0's scheme](https://github.com/nrfconnect/sdk-nrf/blob/master/subsys/bootloader/Kconfig#L162), which uses this `<bootloader>_MIN_PARTITION_SIZE` define.

> It's is already possible to do:
>
> CONFIG_PM_PARTITION_SIZE_MCUBOOT=0x7c00
>
> in the prj_minimal.conf
>
> And then just keep 0xc000 as default if nothing is specified.

Yes but this will automatically handle sizing for general, nRF5340, nRF9160, and external crypto, without further input from the user. Isn't this streamlining a good thing for ease of use and future-proofing?

boot/zephyr/Kconfig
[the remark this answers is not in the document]
nvlsianpu: Why this is not aligned to 0x1000? what part of code is posted from 0x3e00? Is it SPM?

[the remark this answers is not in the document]
kyberdin: Partition manager has an `mcuboot_pad` partition of size `0x200` that it places before the mcuboot primary.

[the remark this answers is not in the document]
nvlsianpu: ^^ I forgot about that

nvlsianpu:
DNM before #143

boot/zephyr/Kconfig
kyberdin: Broke this prompt up into chunks that are dependent upon the relevant Kconfig options. Hopefully this is more in line with what you suggest here.

kyberdin:
Latest push should fix the CI build issues detected in the associated [sdk-nrf #3508](https://github.com/nrfconnect/sdk-nrf/pull/3805).

boot/zephyr/Kconfig
```diff
@@ -28,6 +27,24 @@ partition=MCUBOOT_PAD
 partition-size=0x200
 source "${ZEPHYR_BASE}/../nrf/subsys/partition_manager/Kconfig.template.partition_size"
 
+config PM_PARTITION_SIZE_MCUBOOT
```
nvlsianpu, line 30: These multiple definition of same property is not the most elegant.
I would see here the one, maybe using additional modified Kconfig template.
This might be reworked later,

nvlsianpu (approved):

boot/zephyr/Kconfig
tejlmand: > Isn't this streamlining a good thing for ease of use and future-proofing?

Streamlining is always a good thing, but adding yet another symbols that users must consider whether it should be true or false doesn't necessarily make it easier to understand the system, especially not when there is already a setting for handling the size .

Secondly, setting `CONFIG_BOOT_MIN_PARTITION_SIZE=y` doesn't mean this:
`Build for a minimal image size`, it means, set the default value of `PM_PARTITION_SIZE_MCUBOOT` to a minimal size.

It's still the actual value of `PM_PARTITION_SIZE_MCUBOOT` that decides the size, and not `CONFIG_BOOT_MIN_PARTITION_SIZE=y` so the name and text for the symbol is misleading.

> without further input from the user. 

User still needs to build with: `-DCONF_FILE=prj_minimal.conf` so I don't see how having `CONFIG_PM_PARTITION_SIZE_MCUBOOT=0x7c00` instead of `CONFIG_BOOT_MIN_PARTITION_SIZE=y` is easier for the user.
Could you elaborate a bit on the use-case you have in mind where the proposed change is actually making it easier fir the user ? Maybe that could even lead to an improved outcome.

tejlmand: @hakonfam the Kconfig symbol that @kyberdin refers to in this commit: https://github.com/nrfconnect/sdk-nrf/commit/3ad0ab5c25b5b466f0939094fb26ef85e08cd1c3 is actually poorly named, as it doesn't ensure building for minimal size.

It only sets a different default value, it doesn't ensure that minimum size is used.
```
config B0_MIN_PARTITION_SIZE
	bool "Use minimimum partition size"

....
config PM_PARTITION_SIZE_B0_IMAGE
	....
	default 0x8000 if !B0_MIN_PARTITION_SIZE
	....
	default 0x7000
```

kyberdin: @tejlmand Sorry for taking so long to get back to this. After more thought, I agree that `CONFIG_BOOT_MIN_PARTITION_SIZE=y` is misleading vs specifying via `CONFIG_PM_PARTITION_SIZE_MCUBOOT`. Setting a partition value this way in the `prj_minimal.conf` or `external_crypto.conf` is straightforward and makes sense.

However, a bit of a dilemma by removing this:

Let's say:
* Default partition size is `0xc000` in Kconfig.
* `prj_minimal` sets the partition size to `0x7c00`.
* `external_crypto.conf` sets partition size to `0x3e00`

nRF9160 flash blocks for MCUBoot should be in chunks of at least `FPROTECT_BLOCK_SIZE`, but only Kconfig can set it to this define. A `boards/nrf9160_nrf9160_minimal.conf` could set the partition size to a hardcoded value matching `FPROTECT_BLOCK_SIZE`, but `external_crypto.conf` will override everything regardless.

Considering that this PR aims to enable users to:

1. Build a minimal mcuboot
2. Minimize it further with external crypto

I don't see a way of successfully accomplishing both of these goals for current and future device needs by relying on a hardcoded value in the `external_crypto.conf` fragment.

Therefore, I think the current use case works because:

* It handles the necessary logic across boards
* Defining `CONFIG_PM_PARTITION_SIZE_MCUBOOT=xxxx` is always available to a user fragment for overriding everything, should they want to
* Since `CONFIG_BOOT_MIN_PARTITION_SIZE=y` is set in `prj_minimal.conf`, the user doesn't need to remember or interact with it when building a minimal mcuboot with `-Dmcuboot_CONF_FILE=<path to>/prj_minimal.conf`.

To better clarify the Kconfig option for this, perhaps we could rename `BOOT_MIN_PARTITION_SIZE` to `BOOT_USE_MIN_PARTITION_SIZE`?

Similarly, the Kconfig could be refactored into something like (while trying to avoid ["stuck" configs](https://docs.zephyrproject.org/latest/guides/kconfig/tips.html#stuck-symbols-in-menuconfig-and-guiconfig)):

```
config BOOT_USE_MIN_PARTITION_SIZE
	bool "Produce an MCUboot image that has a minimal partition size."
        help
          Enable this to build MCUboot with the minimum partition size for its 

config PM_PARTITION_SIZE_MCUBOOT
	default FPROTECT_BLOCK_SIZE if (SOC_NRF5340_CPUAPP || SOC_NRF9160)
	default 0x3e00 if BOOT_NRF_EXTERNAL_CRYPTO
	default 0x7c00
	depends on BOOT_MIN_PARTITION_SIZE

config PM_PARTITION_SIZE_MCUBOOT
	hex "Flash space allocated for the MCUboot partition." if !BOOT_USE_MIN_PARTITION_SIZE
	default 0xc000
	help
	  Flash space set aside for the MCUboot partition.
```

boot/zephyr/Kconfig
```diff
@@ -28,6 +27,24 @@ partition=MCUBOOT_PAD
...
+	depends on (BOOT_NRF_EXTERNAL_CRYPTO && BOOT_MIN_PARTITION_SIZE)
+
+config PM_PARTITION_SIZE_MCUBOOT
+	hex
+	default FPROTECT_BLOCK_SIZE if SOC_NRF9160
+	default 0x7c00
+	depends on BOOT_MIN_PARTITION_SIZE
+
+config PM_PARTITION_SIZE_MCUBOOT
+	hex "Flash space allocated for the MCUboot partition."
```
tejlmand, line 43: should this still be user configurable when `BOOT_MIN_PARTITION_SIZE=y` ?

boot/zephyr/Kconfig
```diff
@@ -28,6 +27,24 @@ partition=MCUBOOT_PAD
...
+config PM_PARTITION_SIZE_MCUBOOT
+	hex
+	default FPROTECT_BLOCK_SIZE if (SOC_NRF5340_CPUAPP || SOC_NRF9160)
+	default 0x3e00
+	depends on (BOOT_NRF_EXTERNAL_CRYPTO && BOOT_MIN_PARTITION_SIZE)
+
+config PM_PARTITION_SIZE_MCUBOOT
+	hex
+	default FPROTECT_BLOCK_SIZE if SOC_NRF9160
+	default 0x7c00
```
tejlmand, line 39: Note, this will not be updated if user later sets `BOOT_MIN_PARTITION_SIZE=y` using menuconfig or directly in `.config`.

See more here: https://docs.zephyrproject.org/latest/guides/kconfig/tips.html#stuck-symbols-in-menuconfig-and-guiconfig

boot/zephyr/Kconfig
tejlmand: >  Similarly, the Kconfig could be refactored into something like (while trying to avoid "stuck" configs)

As long as you have a prompt, then you will hit the stuck configs.
https://github.com/kyberdin/sdk-mcuboot/blob/eb69274466d280aedc724a3e6e0f55d1fab56717/boot/zephyr/Kconfig#L43

kyberdin: @tejlmand I have replaced `BOOT_MIN_PARTITION_SIZE` with `BOOT_USE_MIN_PARTITION_SIZE` for the clarity mentioned in my previous comment.

Adding `depends on` to the partition prompt failed to build due to extra tokens, so I used `if !BOOT_USE_MIN_PARTITION_SIZE` in its place.

boot/zephyr/Kconfig
kyberdin: I think this is addressed from [the latest change](https://github.com/nrfconnect/sdk-mcuboot/pull/137#discussion_r612301746).

kyberdin:
@nvlsianpu it looks like the CI for MCUboot is running into the same errors [mentioned in an earlier comment](https://github.com/nrfconnect/sdk-mcuboot/pull/137#issuecomment-770781094). Am I understanding this correctly? Please advise.

tejlmand (approved):

nvlsianpu:
@ChristopherBittner @mswarowsky  ^^
Can we do somethin on these errors?

thst-nordic:
I believe kconfig has cross dependency with other repos now.  It is being worked on. 

If you can verify the kconfig in the sdk-nrf PR this error can be ignored.

mswarowsky:
> @ChristopherBittner @mswarowsky ^^
> Can we do somethin on these errors?

I haven't worked with this CI parts yet, so no good idea, sorry

kyberdin:
Testing this locally works fine for minimal builds until building with the `external_crypto.conf` fragment, which then hangs [here](https://github.com/kyberdin/sdk-mcuboot/blob/minimal-image-size/boot/zephyr/main.c#L554), apparently unable to find a bootable image.

Spoke with @sigvartmh , looks related to #146.

sigvartmh:
Guess this can be merged now?

kyberdin:
> Guess this can be merged now? Maybe wait for CI fix if needed?

Yes, once its [manifest PR](https://github.com/nrfconnect/sdk-nrf/pull/3805) CI has a green light (or at least no failures related to these changes).

tejlmand:
please rebase

kyberdin:
@nvlsianpu checkpatch failed because of a warning about file permissions. Looks similar to one of the earlier comments in this PR.

How do I fix this?

PR #145, opened 2021-03-08
Upmerge 05.03.2021

anangl:


nvlsianpu (approved):

boot/zephyr/prj.conf
```diff
@@ -31,7 +31,7 @@ CONFIG_FPROTECT=y
 # CONFIG_I2C is not set
 
 CONFIG_LOG=y
-CONFIG_LOG_MINIMAL=y
```
MarekPieta, line 34: Shouldn't comment state `former CONFIG_LOG_MINIMAL` ?

nvlsianpu: ^^ yes I should, but this comes from the upstream, my fault.

jfischer-no (approved):

PR #152, opened 2021-05-13
[nrf noup] boot_serial: Add SMP command for erasing settings

sigvartmh:
In the default command group 0 I have added the erase command with ID 6.
Sending this command with operation WRITE or READ will trigger the
command to erase settings flash area.

A successful command will return the following data structure encoded
in cbor:

```
{ rc : <int> }
```

Where `rc` will be the return value of the `erase` function or `open`
function if `open` fails.


sigvartmh:
This PR is now dependent on https://github.com/nrfconnect/sdk-nrf/pull/4537

boot/boot_serial/src/boot_serial.c
```diff
@@ -491,7 +512,13 @@ boot_serial_input(char *buf, int len)
         case NMGR_ID_RESET:
             bs_reset(buf, len);
             break;
+#if defined(CONFIG_SETTINGS)
+        case NMGR_ID_ERASE_STORAGE:
+            bs_erase_storage(buf, len);
+            break;
+#endif
```
de-nordic, line 519: In mgmt/include/mgmt/mgmt.h of mcumgr library there are defined groups of  commands, the one that may be interesting for us here are:
`MGMT_GROUP_ID_CONFIG` or `MGMT_GROUP_ID_PERUSER`
The first one is system level one, the other is user level (so lets say Zephyr level).
We can probably use either of these groups,but lets focus on `MGMT_GROUP_ID_PERUSER`
The groups is user specific, so we could add any commands we like as long as we make it standard for our system (Zephyr). Lets add here command like NMGR_ID_ERASE_SETTINGS (and separately NMGR_ID_ERASE_STORAGE, as these are not always the same).

Next call intermediate function like `imp_system_specific_mgmt_group_id_peruser` providing it with the hdr (or the id) and the other parameters of `bs_erase_storage` and make that function do the job.

Enclose the call in `CONFIG_ENABLE_NMGR_GROUP_ID_PERUSER`, this way systems that would like to implement own stuff of this kind may just enable it and provide own implementation of `imp_system_specific_mgmt_group_id_peruser` or whatever we call this.

Thoughts?

de-nordic (changes requested):

boot/boot_serial/src/boot_serial.c
sigvartmh: So for this specific implementation that won't work since mcuboot does not use mcumgr but I do agree putting the function inside the MGMT_GROUP_ID_PERUSER sounds a bit better than adding an additional command to the OS group. I'll look into how to best integrate it in the way you described.

de-nordic: Yes it does not use library, but you may use mcumgr cli to control it. Although not all things work (on both sides) the ids should be kept in some kind of order, to maintain compatibility. There is risk that someone will add, in mcumgr common code, some id with the same number that does other thing. The PERUSER group will be safer for us.

carlescufi (changes requested):
@sigvartmh please rebase this and then, once reviewed, send it upstream. Thanks!

PR #165, opened 2021-08-18
nRF53 multi-image dual core update

sigvartmh:
Add support for 2 images in `pm.yml` using the `UPDATEABLE_IMAGE` configuration in MCUBoot. Enabling it will create two new partitions called `mcuboot_primary_1` and `mcuboot_secondary_1`. If you enable `flash_simulator` the `primary_1` will be placed in RAM.

This also adds hooks file for nRF53 for handling PCD multi-image updates using the hook interface introduced [here](https://github.com/mcu-tools/mcuboot/pull/1058).


sigvartmh:
I should probably add a hook file for single image use-case for nRF53.

sigvartmh:
Added DNM since I have to update the hook API

boot/zephyr/include/sysflash/sysflash.h
```diff
@@ -16,20 +16,18 @@
...
-#define FLASH_AREA_IMAGE_PRIMARY(x)            \
-        ((x == 0) ?                            \
-           PM_MCUBOOT_PRIMARY_ID :             \
-         (x == 1) ?                            \
-          (uint32_t)_image_1_primary_slot_id : \
+#define FLASH_AREA_IMAGE_PRIMARY(x)          \
+        ((x == 0) ?                          \
+           PM_MCUBOOT_PRIMARY_ID :           \
+         (x == 1) ?                          \
+           PM_MCUBOOT_PRIMARY_1_ID :         \
```
hakonfam, line 23: Doesn't this break existing solution for dual primary single secondary which is used when updating mcuboot?

hakonfam (changes requested):

boot/zephyr/include/sysflash/sysflash.h
sigvartmh: Thought that it wasn't used since I didn't find any reference to it before now `nrf/subsys/bootloader/image/multi_image_mcuboot.conf` I think I forgot to test it properly.

My suggestion is to accept the limitation that you cannot use it with secure boot and #elif this with `CONFIG_SECURE_BOOT`.

hakonfam: Yes I agree, we only need to support a limited use case for this, and not working with `SECURE_BOOT` is not a problem IMO.

hakonfam:
Replaced by https://github.com/nrfconnect/sdk-mcuboot/pull/170

PR #166, opened 2021-08-21
[nrf fromlist] boot: zephyr: Fix misaligned #else for `irq_lock`

sigvartmh:
So in PR #1089 I misaligned the `#else` used for the `irq_lock` as it
should be aligned with `CONFIG_MCUBOOT_CLEANUP_ARM_CORE`. So that
`irq_lock`is called when we don't call `__disable_irq` from arm core
clean up code.


hakonfam (changes requested):
Please add a ref to the upstream PR

nvlsianpu (approved):

nvlsianpu:
@hakonfam I added the reference in the PR description.

PR #157, opened 2021-06-22
[nrf noup] mcuboot: Only include pcd.h if the nRF53 network core is being upgraded

osaether:
Only include pcd.h in mcuboot when CONFIG_NRF53_UPGRADE_NETWORK_CORE is defined

ref: NCSDK-7433

Signed-off-by: Ole Sæther <ole.saether@nordicsemi.no>

PR #167, opened 2021-08-23
[nrf noup] mcuboot: Only include pcd.h if the nRF53 network core is upgraded

nvlsianpu:
Only include pcd.h in mcubot when CONFIG_NRF53_UPGRADE_NETWORK_CORE is
defined

This is reply of #157 which content was disregarded in the last upmerge


mbolivar-nordic (approved):

osaether (approved):
LGTM

PR #170, opened 2021-08-30
Nrf53 simultaneous update

hakonfam:
[DRAFT] Only last commit will be part of this PR


nvlsianpu (commented):
loos fine to me.
Just noted that CONFIG_SECURE_BOOT can't be simultaneously selected with CONFIG_NRF53_MULTI_IMAGE_UPDATE. But this was on of assumption on current development phase.

PR #171, opened 2021-08-31
[nrf fromtree] Allow to boot primary image if secondary one is unreachable

nvlsianpu:
Parent PR: https://github.com/mcu-tools/mcuboot/pull/1038

In case secondary image device has a malfunction,
It might be still possible to boot the primary image.

This make sense especially if the secondary image resides in an
external flash which might be damaged while SoC is still working.

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -381,7 +381,8 @@ boot_read_sectors(struct boot_loader_state *state)
 
     rc = boot_initialize_area(state, FLASH_AREA_IMAGE_SECONDARY(image_index));
     if (rc != 0) {
-        return BOOT_EFLASH;
+        /* We need to differentiate from the primary image issue */
+        return BOOT_EFLASH_SEC;
     }
 
 #if MCUBOOT_SWAP_USING_SCRATCH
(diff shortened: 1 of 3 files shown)


de-nordic (approved):

boot/bootutil/src/loader.c
```diff
@@ -1758,7 +1759,11 @@ boot_prepare_image_for_update(struct boot_loader_state *state,
          * if there is one.
          */
         BOOT_SWAP_TYPE(state) = BOOT_SWAP_TYPE_NONE;
-        return;
+        if (rc == BOOT_EFLASH)
+        {
+            /* Only return on error from the primary image flash */
+            return;
+        }
```
SebastianBoe, line 1766: The if-statement above uses a different coding convention

if
{
}

if {
}

SebastianBoe (changes requested):

boot/bootutil/src/loader.c
SebastianBoe: Already committed upstream.

Not worth the git paperwork to fix.

SebastianBoe (approved):

PR #175, opened 2021-09-08
[nrf fromlist] zephyr: Add multi-image DFU for Thingy:53

MarekPieta:
Upstream PR:
https://github.com/mcu-tools/mcuboot/pull/1136

Change enables multi-image DFU in Thingy:53 configuration.
The network core cannot access external flash directly. The flash
simulator must be used to provide a memory region that is used to
forward the new firmware to the network core.

Jira: NCSDK-10750


boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
```diff
@@ -64,3 +66,11 @@ CONFIG_RESET_ON_FATAL_ERROR=n
 CONFIG_SPI=n
 CONFIG_I2C=n
 CONFIG_UART_NRFX=n
+
+# The following configurations are required to support simultaneous multi image update
+CONFIG_PCD_APP=y
+CONFIG_UPDATEABLE_IMAGE_NUMBER=2
```
tejlmand, line 72: should thingy53 always be doing simultaneously  multi image updates ?

boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
```diff
@@ -64,3 +66,11 @@ CONFIG_RESET_ON_FATAL_ERROR=n
...
+CONFIG_BOOT_UPGRADE_ONLY=y
+CONFIG_FLASH_SIMULATOR=y
+CONFIG_FLASH_SIMULATOR_DOUBLE_WRITES=y
+CONFIG_FLASH_SIMULATOR_STATS=n
```
tejlmand, line 76: Why is flash simulator needed for this ?
https://github.com/nrfconnect/sdk-zephyr/blob/14f880944a81c1ac4796dbf21c325c26b2d06ceb/drivers/flash/Kconfig.simulator#L6

No info is provided in this file and the commit message doesn't describe it either.
Please explain why this is needed.

tejlmand (commented):
General comment.
Why is this a `noup` and not something that should be send upstream.

tejlmand (changes requested):
See review here: https://github.com/nrfconnect/sdk-mcuboot/pull/175#pullrequestreview-750020592

boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
de-nordic: > 
> No info is provided in this file and the commit message doesn't describe it either.
> Please explain why this is needed.

The upload to nrf53 netcore goes through a "window" that the flash simulator provides. There is no direct access to the flash for the netcore update.

tejlmand: Then I request three things regarding this:
1. Provide detailed information in the commit message so that someone wondering why this is added can understand when using `git log`.
2. A comment directly in the prj.conf.
3. I think using `FLASH_SIMULATOR` for such purpose is principally wrong, but accepted for this particular PR if you create a followup task. Even if we are providing / re-using the simulator implementation, I think we should provide dedicated nRF Kconfig settings for this use-case.
    That will make it clearer for our useers, as well as we can provide required selects (maybe even with prompt-less symbols) for a better user-experience.

tejlmand: For example here: https://github.com/nrfconnect/sdk-zephyr/blob/14f880944a81c1ac4796dbf21c325c26b2d06ceb/drivers/flash/Kconfig.nrf

boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
MarekPieta: I think that this should always be supported by the bootloader, because if network core is in use we may need to update it at some point.  Updating only application core is not enough in my opinion.

de-nordic: I agree with @MarekPieta here, multi-image should be default for any device that may required update of additional cores.

boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
MarekPieta: Added explanation in comments and created the requested Jira ticket.

boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
tejlmand: that sounds like something that should be present in nrf5340dk_nrf5340_cpu also then ?

MarekPieta: I am not sure if we are introducing support for this board now. I think the focus is only on Thingy:53.

boot/bootutil/zephyr/CMakeLists.txt
```diff
@@ -16,6 +16,30 @@ zephyr_library_named(mcuboot_util)
...
+
+  if(DEFINED HOOKS_FILE)
+      zephyr_library_sources(
+        ${HOOKS_FILE}
+      )
+  else()
+    message(STATUS "No hooks implementation file.")
+  endif()
+endif()
+
```
tejlmand, line 42: why aren't we just doing:
```
# <ncs>/<path>/CMakeLists.txt
target_sources(mcuboot_util PRIVATE <hook-file.c>)
```
instead of going through KConfig ?

tejlmand (changes requested):
seems overly complex to me.

boot/bootutil/zephyr/CMakeLists.txt
MarekPieta: The code was simply moved from other CMakeLists file. Could we leave it as it is for now?
@nvlsianpu, do you have any input here?

nvlsianpu: @tejlmand This path is needed for the upstream. It also allow to add the hook file to the build somewhere else.

de-nordic: The script may look complicated but it does its job. I think we can optimize it later, and currently we need it. It does not affect size of binary; just makes build take longer.

de-nordic (approved):

boot/bootutil/zephyr/CMakeLists.txt
tejlmand: >  It also allow to add the hook file to the build somewhere else.

Sorry, but i'm not sure I understand what you mean by this ?
What you are requesting is something supported by CMake already, as I showed.

> The code was simply moved from other CMakeLists file. 

Well, it was introduced short time ago, maybe we should do things properly before rolling it out in a release.
https://github.com/mcu-tools/mcuboot/pull/1058

tejlmand: Think I found it: https://github.com/nrfconnect/sdk-nrf/blob/master/modules/mcuboot/hooks/nrf53_hooks.c

nvlsianpu: ^^ yes this is the file we are using.

tejlmand: See alternative proposal which even get rid of the relative / absolute path issue when using Kconfig:
https://github.com/tejlmand/fw-nrfconnect-nrf/commit/b29959dc1f0be337b96403f13c7e4128125f4391
https://github.com/tejlmand/fw-nrfconnect-mcuboot/commit/46c70decc0dd612b2a03ec7a24042fc090f8990e

tejlmand: in that alternative we can discuss which CMakeLists.txt file is the most appropriate location for the `target_sources()` code.

boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
tejlmand: @Rallare could you please take a look here.

boot/bootutil/zephyr/CMakeLists.txt
nvlsianpu: @tejlmand  what you suggested looks fine to me. Path here is merged in the upstream now. What i want to have upstream is possibility for having multiple build configuration - for having test-build with hooks as well. CMake is not enough to make project configurable (can you propose something else I can use?).
Anyway - User which will use target_source() wil just set `CONFIG_BOOT_IMAGE_ACCESS_HOOKS_FILE=""`, so this patch will does nothing in that case.

boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
Rallare: I agree with the fundamental thought here, which is that the appcore needs a "ramdisk" storage for the net-core's binary. This is how the hardware expects sharing of data, so we need to work with it. If you add this to the overall description, it is much easier to understand why it's done.

Is the configurations exposed to users, both cmake and kconfig, in this PR expected to change in the future? 
If yes; add a temp-hack comment on it. It is important that we do not merge anything that we already know will be re-named or even re-written in the near future.

boot/bootutil/zephyr/CMakeLists.txt
tejlmand: > @tejlmand what you suggested looks fine to me. Path here is merged in the upstream now. What i want to have upstream is possibility for having multiple build configuration - for having test-build with hooks as well. CMake is not enough to make project configurable (can you propose something else I can use?).

This is what we do everywhere in Zephyr, we select an option and based on that option CMake determines the file to compile. We never (and should never) point to a c-file from Kconfig.

For example for GPIO, a user can enable GPIO and select a driver:
https://github.com/zephyrproject-rtos/zephyr/blob/e8620a7416e4d51285fe5509e3a5ccbaaaba2818/drivers/gpio/Kconfig.nrfx#L4
```
menuconfig GPIO_NRFX
	bool "nRF GPIO driver"
```
and that results in `gpio_nrfx.c` to be included in the lib.
https://github.com/zephyrproject-rtos/zephyr/blob/e8620a7416e4d51285fe5509e3a5ccbaaaba2818/drivers/gpio/CMakeLists.txt#L21
```
zephyr_library_sources_ifdef(CONFIG_GPIO_NRFX       gpio_nrfx.c)
```

so if mcuboot supports native hook, I would suggest those to be available in same way as we do everywhere else, and not invent a new scheme for that.

Then a downstream user can follow the same principle for adding out-of-tree hooks, like is done for entropy.
https://github.com/nrfconnect/sdk-nrf/blob/f10756d98c5bec203971771d8700ad81aa0283d4/drivers/entropy/Kconfig#L7-L8
```
config ENTROPY_CC3XX
	bool "Arm CC3XX RNG driver for Nordic devices" if !BUILD_WITH_TFM
```
which results in:
https://github.com/nrfconnect/sdk-nrf/blob/f10756d98c5bec203971771d8700ad81aa0283d4/drivers/entropy/CMakeLists.txt#L6-L7
```
zephyr_library_amend()
zephyr_library_sources_ifdef(CONFIG_ENTROPY_CC3XX entropy_cc310.c)
```
or they can resort to plain cmake if they like.

> Anyway - User which will use target_source() wil just set CONFIG_BOOT_IMAGE_ACCESS_HOOKS_FILE="", so this patch will does nothing in that case.

Besides not following the pattern we use everywhere else, then it will also result in this message without any guidance on what might be wrong:
```
  else()
    message(STATUS "No hooks implementation file.")
  endif()
```

nvlsianpu: I think you are right @tejlmand. Will register this as issue upstream an will consult with you regard the solution when will I will be fixing that. As far I understand I can just remove whole complex patch resolution from the CMake and add proper hint to the Kconfig help. For doing upstream test-build i can add hooks_saple.c in upstream camake depending on a build commanding option.

tejlmand: as soon as such issue has been created, post a link and I'll approve this PR, as I don't believe I have any other blocking comments.

MarekPieta:
@de-nordic, @nvlsianpu, @tejlmand please rereview

boot/bootutil/zephyr/CMakeLists.txt
nvlsianpu: Marek removed this commit as flash_nop_driver was enough to mitigate the issue which hooks amendment tried to fix.

nvlsianpu (approved):

de-nordic (approved):

hakonfam (approved):

pdunaj (approved):

MarekPieta:
Introduced fix for MCUboot serial recovery image list when all 4 images are flashed. @nvlsianpu marked it as safe to add.

boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
MarekPieta: I fully understood that the configuration is actually needed for nRF5340 hardware in general.

tejlmand: > It is important that we do not merge features that we already know will be re-named or even re-written in the near future without explicitly stating so.

I completely agree on this, it's a hidden temphack :disappointed: . Unfortunately this has already been merged upstream: https://github.com/mcu-tools/mcuboot/pull/1136
meaning the problem exists upstream (and must be cleaned up upstream).
@MarekPieta when a commit is merged upstream, the commit title should be changed to `[nrf fromtree]`.

MarekPieta: The configuration may be improved in the future, but I would prefer to have a common MCUboot configuration that would be used for Thingy:53 apps now (so we must use existing configs).
For example the flash simulator name may actually change in the future (it was already requested)

tejlmand (approved):
approving under the expectation that all identified and outstanding issues will be addressed and improved before next released.

shantha-14 (approved):

pdunaj:
Hi @carlescufi , can you merge this one?

boot/zephyr/boards/thingy53_nrf5340_cpuapp.conf
MarekPieta: @tejlmand, using fromtree now. The PRs were merged in meantime.

PR #181, opened 2021-11-07
[nrf noup] zephyr: Added TF-M present Kconfig

Vge0rge:
-Add a configuration to allow the MCUBOOT be
 aware that the TF-M will be present in the build.
 This is needed since FW_INFO is not available for
 TF-M and MCUBOOT enforces the use of the EXT_APIs.

Ref: NCSDK-12021

Signed-off-by: Georgios Vasilakis <georgios.vasilakis@nordicsemi.no>

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -89,6 +89,12 @@ config BOOT_USE_NRF_EXTERNAL_CRYPTO
 	help
 	  Use Shared crypto for crypto primitives.
 
+config MCUBOOT_TFM_PRESENT
+	bool "TF-M is present"
+	help
+	  This option allows MCUBOOT perform configurations to allow
+	  compatibility with TF-M.
+
 menu "MCUBoot settings"
 
 config SINGLE_APPLICATION_SLOT


sigvartmh (approved):

hakonfam (changes requested):
IMO we should try to solve this problem without introducing new Kconfig options. If we do need to add a new kconfig option please add this option to the mcuboot module kconfig in the sdk-nrf repo. See https://github.com/nrfconnect/sdk-nrf/blob/main/modules/Kconfig.mcuboot 

That way we don't need downstream patches and multi repo PRs.

Vge0rge:
> IMO we should try to solve this problem without introducing new Kconfig options. If we do need to add a new kconfig option please add this option to the mcuboot module kconfig in the sdk-nrf repo. See https://github.com/nrfconnect/sdk-nrf/blob/main/modules/Kconfig.mcuboot
> 
> That way we don't need downstream patches and multi repo PRs.
 @hakonfam 
Since the fw_info configurations are using Kconfig I am not sure that I know any clean way of doing this without using a new Kconfig option. MCUBOOT AFAIK does now know if the app it will load is TF-M or a regular app. I am closing this PR since the Kconfig.mcuboot is for sure a better placment for this and I updated the sdk-nrf PR with this.

PR #183, opened 2021-11-10
[nrf noup] kconfig: reduce downstream patch

hakonfam:
Source the file 'Kconfig.mcuboot' from sdk-nrf to reduce the amount
of downstream patches in boot/zephyr/Kconfig.

Ref: NCSDK-11177

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -8,6 +8,8 @@ mainmenu "MCUboot configuration"
 
 comment "MCUboot-specific configuration options"
 
+source "$(ZEPHYR_NRF_MODULE_DIR)/modules/mcuboot/boot/zephyr/Kconfig"
+
 # Hidden option to mark a project as MCUboot
 config MCUBOOT
 	default y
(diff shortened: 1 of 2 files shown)


nvlsianpu (commented):
Nice!
What about adding names to choices we had modified so it will be possible to move downstream patches nrf?

hakonfam:
> Nice! What about adding names to choices we had modified so it will be possible to move downstream patches nrf?

I didn't know that this was possible, I will try to do that.

boot/zephyr/Kconfig
```diff
@@ -16,35 +16,6 @@ config MCUBOOT
 	select USE_DT_CODE_PARTITION if HAS_FLASH_LOAD_OFFSET
 	select MCUBOOT_BOOTUTIL_LIB
 
-config BOOT_USE_MIN_PARTITION_SIZE
```
tejlmand, line 19: this is introduced in earlier commit.
Should we do a revert and then re-commit so that the patchset can be reduced in future history rewrite ?

Or make a comment in the commit message that it should be squashed with commit-x.

boot/zephyr/Kconfig
```diff
@@ -766,4 +728,5 @@ config MCUBOOT_BOOTUTIL_LIB_OWN_LOG
 	bool
 	default n
 
+source "$(ZEPHYR_NRF_MODULE_DIR)/modules/Kconfig.mcuboot"
```
tejlmand, line 731: :+1: for the pronciple.

Question: should this be first so that we have the possibility to overwrite any mcuboot defaults in our custom nrf `Kconfig.mcuboot`.

tejlmand (commented):
Thanks for this.

Principle looks good.

boot/zephyr/Kconfig
hakonfam: I moved it to the first line now 👍🏻

boot/zephyr/Kconfig
hakonfam: I will identify the commits that added these and try to come up with a squashing-scheme

hakonfam:
@nvlsianpu I added names to the choices as you suggested, so now there should be no dowstream patches for this file.

boot/zephyr/Kconfig
hakonfam: I'm getting conflicts when reverting due to how the last upmerge was handled. IMO we could handle this by informing those responible for the next upmerge that none of the "noup" should have any modifications to `boot/zephyr/Kconfig` and that this commit should be removed at that point.

nvlsianpu (approved):

boot/zephyr/Kconfig
```diff
@@ -74,18 +47,15 @@ config BOOT_USE_CC310
...
 
-config BOOT_USE_NRF_EXTERNAL_CRYPTO
+config NRFXLIB_CRYPTO
+	bool
+	default n
+
+config NRF_CC310_BL
 	bool
-	# Hidden option
 	default n
```
tejlmand, line 58: Why us there a `default n` here ?

boot/zephyr/Kconfig
```diff
@@ -74,18 +47,15 @@ config BOOT_USE_CC310
 
 config BOOT_USE_NRF_CC310_BL
 	bool
-	select NRF_CC310_BL
+	default n
```
tejlmand, line 50: Why is there a `default n` here ?

nvlsianpu: this patch reverts this line to its upstream state

boot/zephyr/Kconfig
nvlsianpu: this patch reverts this line to its upstream state

tejlmand (approved):

PR #184, opened 2021-11-19
[nrf fromlist] boo_serial: cleanup max output size calculation

nvlsianpu:
Calculation of BOOT_SERIAL_OUT_MAX was based on MCUBOOT_IMAGE_NUMBER
while in other places BOOT_IMAGE_NUMBER is taken int account as
images number. Let's align the calculation to use same literal value
as others.


tejlmand:
Just noticed, the commit title says:
> boo_serial

not
> boot_serial

nvlsianpu:
DNM as this will go within #183

PR #185, opened 2021-11-29
[nrf noup] pm.yml: reserve area for swapping

hakonfam:
Moving using swap requires one page to work.
Currently there is nothing stopping an image from using all pages
in its partition, which would break the swapping mechanism.

Fix this by introducing a placeholder partition with the size
of a page.

Also remove unnecessary alignment of mcuboot secondary.

Ref: NCSDK-7203


mbolivar-nordic:
Please rebase onto the new history in main

tejlmand (approved):
lgtm, but please rebase.

boot/zephyr/pm.yml
```diff
@@ -24,7 +33,6 @@ mcuboot_secondary:
     align: {start: 4}
 #else
   placement:
-    align: {start: CONFIG_FPROTECT_BLOCK_SIZE}
```
MarekPieta: Should we also remove the `align_next`? Should we consider aligning to size of the FLASH page that is the smallest chunk of memory which could be erased separately?

boot/zephyr/pm.yml
```diff
@@ -13,6 +13,15 @@ mcuboot_primary_app:
 mcuboot_primary:
   span: [mcuboot_pad, mcuboot_primary_app]
 
+#if defined(CONFIG_BOOT_SWAP_USING_MOVE)
+# This region is used to ensure that at least 1 page remains unused
+mcuboot_swap_area:
```
MarekPieta, line 18: Can `mcuboot_secondary_app` still share size with `mcuboot_primary_app`?

hakonfam: yes that should still work.

boot/zephyr/pm.yml
hakonfam: > Should we also remove the `align_next`? 

Do we need that? We need the start of that to be aligned with FPROTECT granularity, otherwise we cannot safely separate them.

> Should we consider aligning to size of the FLASH page that is the smallest chunk of memory which could be erased separately?

Do you mean the alignment of the `align_next` ? Or the `swap area` ?

de-nordic (approved):

boot/zephyr/pm.yml
MarekPieta: Sorry, I confused this one with swap using scratch method initially. Now I see it's needed.

MarekPieta: I am just not sure if it should be in `mcuboot_primary_app`, maybe it would be more intuitive to put it in `mcuboot_primary` (since it's not part of application I think). Still, it would work anyway, so I do not block.

boot/zephyr/pm.yml
MarekPieta: > Do we need that? We need the start of that to be aligned with FPROTECT granularity, otherwise we cannot safely separate them.

If we want to align the partition to FPROTECT granularity, shouldn't we also align the partition start here? Or it's enough that the alignment is added under `mcuboot_secondary_pad`?

MarekPieta: > Do you mean the alignment of the `align_next` ? Or the `swap area` ?
I was thinking about ensuring that image slots' size is multiple of `CONFIG_FLASH_LAYOUT_PAGE_SIZE` to ensure that we can erase it safely without interfering with other data in FLASH (if we place the image data in internal FLASH).

boot/zephyr/pm.yml
hakonfam: Hm, I see your point. Actually, we might end up signing a bunch of 0xFF if we put this inside `mcuboot_primary_app` (havent double checked that), I'll make a fix.

nordicjm (approved):

boot/zephyr/pm.yml
```diff
@@ -13,6 +13,15 @@ mcuboot_primary_app:
...
+  size: CONFIG_FLASH_LAYOUT_PAGE_SIZE
+  placement:
+    after: app
+  inside: mcuboot_primary_app
```
MarekPieta, line 22: Are you going to move it to the `mcuboot_primary` area (as we discussed) ?

hakonfam: Its a while since we discussed this, what was the reason for placing it there instead of inside `mcuboot_primary_app` - it should have the same result no?

PR #187, opened 2021-12-02
[nrf noup] zephyr/nrf_cleanup: clean-up PPI/DPPI

nvlsianpu:
Disable all ppi/dppi channels in case it were used.

fix for NCSDK-12718

### DIFF
--- a/boot/zephyr/nrf_cleanup.c
+++ b/boot/zephyr/nrf_cleanup.c
@@ -11,6 +11,22 @@
 #if defined(NRF_RTC0) || defined(NRF_RTC1) || defined(NRF_RTC2)
     #include <hal/nrf_rtc.h>
 #endif
+#if defined(NRF_PPI)
+    #include <hal/nrf_ppi.h>
+#endif
+#if defined(NRF_DPPIC)
+    #include <hal/nrf_dppi.h>
+#endif
+
+#include <string.h>
+
+#define NRF_UARTE_SUBSCRIBE_CONF_OFFS offsetof(NRF_UARTE_Type, SUBSCRIBE_STARTRX)
+#define NRF_UARTE_SUBSCRIBE_CONF_SIZE (offsetof(NRF_UARTE_Type, EVENTS_CTS) -\
+                                       NRF_UARTE_SUBSCRIBE_CONF_OFFS)
+
+#define NRF_UARTE_PUBLISH_CONF_OFFS offsetof(NRF_UARTE_Type, PUBLISH_CTS)
+#define NRF_UARTE_PUBLISH_CONF_SIZE (offsetof(NRF_UARTE_Type, SHORTS) -\
+                                     NRF_UARTE_PUBLISH_CONF_OFFS)
 
 #if defined(NRF_RTC0) || defined(NRF_RTC1) || defined(NRF_RTC2)
 static inline void nrf_cleanup_rtc(NRF_RTC_Type * rtc_reg)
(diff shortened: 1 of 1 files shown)


boot/zephyr/nrf_cleanup.c
```diff
@@ -11,6 +11,13 @@
 #if defined(NRF_RTC0) || defined(NRF_RTC1) || defined(NRF_RTC2)
     #include <hal/nrf_rtc.h>
 #endif
+#if defined(NRF_PPI)
+    #include <hal/nrf_ppi.h>
+#endif
+#if defined(NRF_DPPI)
```
anangl, line 17:
```suggestion
#if defined(NRF_DPPIC)
```

anangl (changes requested):

anangl (approved):

boot/zephyr/nrf_cleanup.c
```diff
@@ -44,6 +51,12 @@ void nrf_cleanup_peripheral(void)
 #if defined(NRF_UARTE1)
     nrf_uarte_disable(NRF_UARTE1);
     nrf_uarte_int_disable(NRF_UARTE1, 0xFFFFFFFF);
+#endif
+#if defined(NRF_PPI)
+nrf_ppi_channels_disable_all(NRF_PPI);
```
balaji-nordic, line 56: Please fix the alignment here and in line 59

balaji-nordic (changes requested):
Tested the changes. They work fine. Just some alignment fixes needed.

anangl (changes requested):

boot/zephyr/nrf_cleanup.c
```diff
@@ -44,6 +51,12 @@ void nrf_cleanup_peripheral(void)
 #if defined(NRF_UARTE1)
     nrf_uarte_disable(NRF_UARTE1);
     nrf_uarte_int_disable(NRF_UARTE1, 0xFFFFFFFF);
+#endif
+#if defined(NRF_PPI)
+    nrf_ppi_channels_disable_all(NRF_PPI);
+#endif
+#if defined(NRF_DPPIC)
+    nrf_dppi_channels_disable_all(NRF_DPPIC);
```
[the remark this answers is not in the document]
nvlsianpu: I would add such "register reset" only for uarte.
I think for another peripheral (for example SPIM, QSPI) it should be done in other PR as this should go quickly as the bugfix now.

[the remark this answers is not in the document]
balaji-nordic: Am I right to assume that we only need to clean up the peripherals used by bootloader?

boot/zephyr/nrf_cleanup.c
```diff
@@ -40,10 +48,28 @@ void nrf_cleanup_peripheral(void)
 #if defined(NRF_UARTE0)
     nrf_uarte_disable(NRF_UARTE0);
     nrf_uarte_int_disable(NRF_UARTE0, 0xFFFFFFFF);
+#if defined(NRF_DPPIC)
+    /* Clear all SUBSCRIBE configurations. */
+    memset((uint8_t *)NRF_UARTE0 + 0x80, 0, 0x80);
```
balaji-nordic, line 53: Since we are assuming the location of SUBSCRIBE and PUBLISH configuration is the same irrespective of the target SoC, I suggest we future proof this code with sufficient BUILD_ASSERTS. For example, we can guard this memset by the following two BUILD_ASSERTs. Similar BUILD_ASSERTs can be added for the other memsets too.



    BUILD_ASSERT((uint8_t *)NRF_UARTE0 + 0x80 == (uint8_t*)&NRF_UARTE0->SUBSCRIBE_STARTRX,
        "Location of UART subscribe configuration is not known for current SoC.");
    BUILD_ASSERT((uint8_t *)NRF_UARTE0 + 0x80 + 0x80 == (uint8_t*)&NRF_UARTE0->EVENTS_CTS,
        "Location of UART subscribe configuration is not known for current SoC.");

I also suggest using macros instead of magic numbers here. 
eg) 
`
#define SUBSCRIBE_CONFIG_OFFSET 0x80 /* Offset of the memory location of the first SUBSCRIBE register. */ 

#define SUBSCRIBE_CONFIG_SIZE 0x80 /* Total memory size occupied by all SUBSCRIBE registers of the UART peripheral. */
`

balaji-nordic:
Suggested rewording of the commit message:

> Cleanup PPI/DPPI channel configuration
> 
> Disable all PPI channels. Clear all subscribe and publish configurations of UARTE0 and UARTE1. 
>
> Fixes NCSDK-12718

balaji-nordic:
Tested the changes again and I confirm that these fix the uart tx errors without additional KConfig changes in the application.

boot/zephyr/nrf_cleanup.c
```diff
@@ -40,10 +48,28 @@ void nrf_cleanup_peripheral(void)
...
+    /* Clear all PUBLISH configurations. */
+    memset((uint8_t *)NRF_UARTE0 + 0x180, 0, 0x80);
+#endif
 #endif
 #if defined(NRF_UARTE1)
     nrf_uarte_disable(NRF_UARTE1);
```
balaji-nordic, line 59: Any idea why mcuboot uses UARTE1?

balaji-nordic (changes requested):

jhn-nordic:
This fix seems to improve reliability of the uart a lot when doing modem tracing. Should be part of the release

jhn-nordic (approved):

boot/zephyr/nrf_cleanup.c
nvlsianpu: better now?

boot/zephyr/nrf_cleanup.c
nvlsianpu: can be used for logging or serial recovery (depends on configuration)

boot/zephyr/nrf_cleanup.c
balaji-nordic: yes. Thanks.

balaji-nordic (approved):

nvlsianpu:
@balaji-nordic I only redacted the commit message as you suggested.

PR #196, opened 2022-02-10
[nrf noup] fixed check of reset addres for the net app

nvlsianpu:
The primary slot for the network core is emulated in RAM.
Its flash_area hasn't got relevant boundaries.

Therefore need to override its boundaries for the check.

ref. NCSDK-13532

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -804,7 +804,24 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
             goto out;
         }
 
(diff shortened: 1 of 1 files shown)


nvlsianpu (commented):
@hakonfam @sigvartmh 
Is it possible to have right offset autogenerated?

nvlsianpu:
> @hakonfam @sigvartmh
Is it possible to have right offset autogenerated?

This will require PM changes - Better to fix this in hacky way for the release.

pdunaj (approved):

hakonfam (approved):

sigvartmh (approved):

PR #197, opened 2022-02-14

chrta:
When mcuboot_secondary is on external flash, the image header cannot
dircetly be accessed via secondary_fa->fa_off. Instead the provided
function boot_img_hdr() is used now.
Additionally a similar issue is present when trying to read the address
of the reset handler. For this flash_area_read() is used now.

With this patch is possible to have the update partiton
mcuboot_secondary on external flash and update a updatable
bootloader (mcuboot) in s0 and/or s1.

<b>Objective: To allow update the mcuboot from the image in the external memory, </b>


chrta:
Opened this pr as discussed in discord with @nvlsianpu 
Feel free to modify this. We use this patch to be able to update the updateable bootloader on a nrf9160.

boot/bootutil/src/loader.c
```diff
@@ -909,9 +909,7 @@ boot_validated_swap_type(struct boot_loader_state *state,
 #if defined(PM_S1_ADDRESS) || defined(CONFIG_SOC_NRF5340_CPUAPP)
     const struct flash_area *secondary_fa =
         BOOT_IMG_AREA(state, BOOT_SECONDARY_SLOT);
-    struct image_header *hdr = (struct image_header *)secondary_fa->fa_off;
-    uint32_t vtable_addr = 0;
-    uint32_t *vtable = 0;
+    struct image_header *hdr = boot_img_hdr(state, BOOT_SECONDARY_SLOT);
```
nvlsianpu, line 912: This seams to be OK as image headers were fetch already to the `state` structure before call to the boot_validate_swap_type() in lines:
https://github.com/nrfconnect/sdk-mcuboot/blob/17f6d045a10a8d2ba9f3dede820e9c6dd81a5c56/boot/bootutil/src/loader.c#L1827 and https://github.com/nrfconnect/sdk-mcuboot/blob/17f6d045a10a8d2ba9f3dede820e9c6dd81a5c56/boot/bootutil/src/loader.c#L1858

jimmyzumthurm:
This patch will make it impossible to update the network core on nrf5340, since pcd_network_core_update is called a few lines under and requires that vtable is defined, which is not the case here.

chrta:
> This patch will make it impossible to update the network core on nrf5340, since pcd_network_core_update is called a few lines under and requires that vtable is defined, which is not the case here.

You are right, i missed that. It seems that `pcd_network_core_update()` only works when the update data is in internal flash or in ram. Any ideas how to solve this issue here? I could recalculate the vtable variable in line https://github.com/nrfconnect/sdk-mcuboot/pull/197/files#diff-a004fdf5907914a360ce1e7da94221c0165d1c4c8ba7b8c4eaf67bd63af395c1R972 (so basically copying the old logic) but i do not have any hardware to check that.

jimmyzumthurm:
@chrta I removed lines 965-991 completely and now I can do DFU of both core when they are in external flash.
The pcd_network_core_update() is now instead called using hooks inside nrf53_hooks.c file. I think it is how it's supposed to work and these lines could be removed from this file.

hakonfam:
Thank you @chrta for this patch. We have added this issue to our internal issue tracking and will make sure its fixed prior to the next release. If you want we can use this PR to get the fix through, or we can create a separate one. You will be credited regardless.

chrta:
@hakonfam Great, thank you. Just do whatever is easier for you. Please let me know if there is anything for me to do.

jimmyzumthurm:
> @chrta I removed lines 965-991 completely and now I can do DFU of both core when they are in external flash. The pcd_network_core_update() is now instead called using hooks inside nrf53_hooks.c file. I think it is how it's supposed to work and these lines could be removed from this file.

@nvlsianpu  @hakonfam  Could you confirm that the chunk of code from line 965 to 991 is outdated since nrf53_hooks.c has been added and that it should be removed ? If it is the case, should I make a PR about it ?

nvlsianpu:
@jimmyzumthurm Hooks do not cover the code you mentioned yet.

hakonfam:
We were not able to push to your remote so we moved the work here https://github.com/nrfconnect/sdk-mcuboot/pull/210.

PR #208, opened 2022-07-25
[nrf noup] bootutil: loader: fix a hardfault when the external second…

aiminhua:
…ary address greater than 1MB

hdr->ih_magic is an internal Flash address though the secondary slot is on external Flash. Without
this fix, hdr->ih_magic would cause a hardfault if the external address is greater than 1MB.
In addition, this fix also moves network_core_update() to nrf53_hooks.c to have a better consistency.

Signed-off-by: Kevin Ai <kevin.ai@nordicsemi.no>


de-nordic:
Rebase needed here.

PR #209, opened 2022-08-03
Upmerge 08 03 2022

nordicjm:


nvlsianpu (commented):
Upmerge is done properly.

nvlsianpu (approved):

PR #210, opened 2022-08-26

osaether:
When mcuboot_secondary is on external flash, the image header cannot
dircetly be accessed via secondary_fa->fa_off. Instead the provided
function boot_img_hdr() is used now.
Additionally a similar issue is present when trying to read the address
of the reset handler. For this flash_area_read() is used now.

With this patch is possible to have the update partiton
mcuboot_secondary on external flash and update a updatable
bootloader (mcuboot) in s0 and/or s1.

Signed-off-by: Ole Sæther <ole.saether@nordicsemi.no>

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -917,10 +917,9 @@ boot_validated_swap_type(struct boot_loader_state *state,
 #if defined(PM_S1_ADDRESS) || defined(CONFIG_SOC_NRF5340_CPUAPP)
     const struct flash_area *secondary_fa =
         BOOT_IMG_AREA(state, BOOT_SECONDARY_SLOT);
-    struct image_header *hdr = (struct image_header *)secondary_fa->fa_off;
-    uint32_t vtable_addr = 0;
-    uint32_t *vtable = 0;
+    struct image_header *hdr = boot_img_hdr(state, BOOT_SECONDARY_SLOT);
     uint32_t reset_addr = 0;
+    int rc = 0;
     /* Patch needed for NCS. Since image 0 (the app) and image 1 (the other
      * B1 slot S0 or S1) share the same secondary slot, we need to check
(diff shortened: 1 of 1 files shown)


sigvartmh (approved):

boot/bootutil/src/loader.c
```diff
@@ -930,16 +929,19 @@ boot_validated_swap_type(struct boot_loader_state *state,
...
+            return BOOT_SWAP_TYPE_FAIL;
+        }
 #ifdef PM_S1_ADDRESS
 #ifdef PM_CPUNET_B0N_ADDRESS
         if(reset_addr < PM_CPUNET_B0N_ADDRESS)
 #endif
         {
             const struct flash_area *primary_fa;
-            int rc = flash_area_open(flash_area_id_from_multi_image_slot(
+        rc = flash_area_open(flash_area_id_from_multi_image_slot(
```
nvlsianpu, line 944: alignment

nvlsianpu (approved):
just one nit on alignment

boot/bootutil/src/loader.c
gmarull: @osaether can you please fix this? We are about to merge for 2.1.0 rc1

gmarull:
merging, last force push is a space change and CI was green before (it seems like a CI script problem)

carlescufi:
@chrta this patch made CI fail so we reverted it here: https://github.com/nrfconnect/sdk-mcuboot/pull/211

@osaether @nvlsianpu can you please give more details as to why this needed reverting?

nvlsianpu:
This patch broke a Matter user-case so we decided to postpone it until next release.

chrta:
@carlescufi @nvlsianpu  Thx for the info

richesonk:
Is the work to fix and submit this patch "after the release" underway and/or is there a way to track if/when that work begins?

PR #218, opened 2022-11-03
[nrf noup] Skip updating sec counters

Vge0rge:
This adds the code and logic to support hardware
counters for rollback protection in MCUBOOT
(CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION).

Ref: NCSDK-9045

Signed-off-by: Georgios Vasilakis <georgios.vasilakis@nordicsemi.no>

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -2296,9 +2296,14 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
         }
 #endif /* MCUBOOT_VALIDATE_PRIMARY_SLOT */
 
+#ifdef PM_S1_ADDRESS
+        if (!image_validated_by_nsib)
+#endif
+        {
         rc = boot_update_hw_rollback_protection(state);
         if (rc != 0) {
             goto out;
+        }
         }
 
         rc = boot_add_shared_data(state, BOOT_PRIMARY_SLOT);
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -2269,19 +2269,19 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
 
 #ifdef MCUBOOT_VALIDATE_PRIMARY_SLOT
 #ifdef PM_S1_ADDRESS
-	/* Patch needed for NCS. Image 1 primary is the currently
-	 * executing MCUBoot image, and is therefore already trusted and
-	 * does not need validation.
-	 */
-    bool image_needs_validation = BOOT_CURR_IMG(state) != 1;
-	if (image_needs_validation)
+        /* Patch needed for NCS. Image 1 primary is the currently
+         * executing MCUBoot image, and is therefore already validated by NSIB and
+         * does not need to also be validated by MCUBoot.
+         */
+        bool image_validated_by_nsib = BOOT_CURR_IMG(state) == 1;
+        if (!image_validated_by_nsib)
 #endif
-	{
+        {
             FIH_CALL(boot_validate_slot, fih_rc, state, BOOT_PRIMARY_SLOT, NULL);
             if (fih_not_eq(fih_rc, FIH_SUCCESS)) {
                 goto out;
             }
-	}
+        }
 #else
         /* Even if we're not re-validating the primary slot, we could be booting
          * onto an empty flash chip. At least do a basic sanity check that


boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
...
+    case 1:
+        err = get_monotonic_counter(COUNTER_DESC_MCUBOOT_HW_COUNTER_ID1, &cur_sec_cnt);
+        break;
+    case 2:
+        err = get_monotonic_counter(COUNTER_DESC_MCUBOOT_HW_COUNTER_ID2, &cur_sec_cnt);
+        break;
+    default:
+        return FIH_FAILURE;
+    }
+
```
SebastianBoe, line 44:
```suggestion
   if(image_id > 2) {
     return FIH_FAILURE;
   }
   
   uint32_t? ids[3] = {
     COUNTER_DESC_MCUBOOT_HW_COUNTER_ID0,
     COUNTER_DESC_MCUBOOT_HW_COUNTER_ID1,
     COUNTER_DESC_MCUBOOT_HW_COUNTER_ID2
  };
               
   err = get_monotonic_counter(ids[image_id], &cur_sec_cnt);
```

For readability and maintainability.

boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
...
+ * Reads the stored value of a given image's security counter.
+ *
+ * @param image_id          Index of the image (from 0).
+ * @param security_cnt      Pointer to store the security counter value.
+ *
+ * @return                  FIH_SUCCESS on success
+ */
+fih_int boot_nv_security_counter_get(uint32_t image_id, fih_int *security_cnt){
+    int err;
+    uint16_t cur_sec_cnt;
```
SebastianBoe, line 24: It's not clear to me why we need this, couldn't we just pass on security_cnt to get_monotonic_counter ?

boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
...
+
+fih_int boot_nv_security_counter_init(void){
+    return FIH_SUCCESS;
+}
+
+/**
+ * Reads the stored value of a given image's security counter.
+ *
+ * @param image_id          Index of the image (from 0).
+ * @param security_cnt      Pointer to store the security counter value.
```
SebastianBoe, line 18:
```suggestion
 * @param[out] security_cnt      Pointer to store the security counter value.
```

boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Updates the stored value of a given image's security counter with a new
+ * security counter value if the new one is greater.
+ *
+ * @param image_id          Index of the image (from 0).
+ * @param img_security_cnt  New security counter value. The new value must be
+ *                          between 0 and UINT32_MAX and it must be greater than
+ *                          or equal to the current security counter value.
+ *
+ * @return                  0 on success; nonzero on failure.
```
SebastianBoe, line 63: Let's be consistent with the error handling.

E.g. we can use this line for the above function as well.

boot/zephyr/nrf_nv_counters.c
SebastianBoe: I see we can do the same refactoring below.

So ids can be put into file-level scope (with static const to ensure it's placed in flash).

boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
...
+        return FIH_FAILURE;
+    }
+
+    if (err != 0) {
+        return FIH_FAILURE;
+    }
+
+    if (counter_value == (uint16_t) img_security_cnt){
+        return FIH_SUCCESS;
+    }
```
SebastianBoe, line 97: Maybe an explanation for this is in order.

e.g.

/*
MCUBoot requires that counter updates to the same value be permitted. But set_monotonic_counter does not permit this, so we check the old value before updating.
*/

Assuming the above is true. I don't know the semantics of set_monotonic_counter

boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
...
+
+    *security_cnt = cur_sec_cnt;
+
+    return FIH_SUCCESS;
+}
+
+/**
+ * Updates the stored value of a given image's security counter with a new
+ * security counter value if the new one is greater.
```
SebastianBoe, line 56:
```suggestion
 * Increases the value of an image's security counter.
```

boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @param image_id          Index of the image (from 0).
+ * @param img_security_cnt  New security counter value. The new value must be
+ *                          between 0 and UINT32_MAX and it must be greater than
```
SebastianBoe, line 60:
```suggestion
 *                          between 0 and UINT16_MAX and it must be greater than
```

Also, maybe be very explicit about whether this includes or excludes 0 and UINT16_MAX. Between is unfortunately ambiguous for some readers.

boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
...
+        break;
+    case 1:
+        err = get_monotonic_counter(COUNTER_DESC_MCUBOOT_HW_COUNTER_ID1, &counter_value);
+        break;
+    case 2:
+        err = get_monotonic_counter(COUNTER_DESC_MCUBOOT_HW_COUNTER_ID2, &counter_value);
+        break;
+    default:
+        return FIH_FAILURE;
+    }
```
SebastianBoe, line 89:
```suggestion
uint32_t old_cnt;
err = boot_nv_security_counter_get(image_id, &old_cnt);
```

boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
...
+    {
+    case 0:
+        return set_monotonic_counter(COUNTER_DESC_MCUBOOT_HW_COUNTER_ID0, (uint16_t) img_security_cnt);
+    case 1:
+        return set_monotonic_counter(COUNTER_DESC_MCUBOOT_HW_COUNTER_ID1, (uint16_t) img_security_cnt);
+    case 2:
+        return set_monotonic_counter(COUNTER_DESC_MCUBOOT_HW_COUNTER_ID2, (uint16_t) img_security_cnt);
+    default:
+        return FIH_FAILURE;
+    }
```
SebastianBoe, line 109: Needs the same refactoring as above.

boot/zephyr/pm.yml
```diff
@@ -85,3 +85,9 @@ mcuboot_secondary_1:
   size: CONFIG_NRF53_RAM_FLASH_SIZE
 
 #endif /* CONFIG_NRF53_MULTI_IMAGE_UPDATE */
+
+#if (CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION)
+provision:
+  size: CONFIG_PM_PARTITION_SIZE_PROVISION
+  region: otp
+#endif /* CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION */
```
SebastianBoe, line 93: Hi, why is this needed?

I would think that when

CONFIG_SECURE_BOOT_STORAGE is enabled this code would already be added though this mechanism:

https://github.com/nrfconnect/sdk-nrf/pull/9196/files#diff-f434425bbc55a28763ba74d09be74c00824ba6d3f768af979614867b3485ec54R81

Perhaps the Kconfig is disabled?

Shall we introduce a new Kconfig that adds pm.yml.secure_boot_storage to the build?

Or should the Kconfig be enabled?

boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
...
+        break;
+    default:
+        return FIH_FAILURE;
+    }
+
+    if (err != 0) {
+        return FIH_FAILURE;
+    }
+
+    if (counter_value == (uint16_t) img_security_cnt){
```
SebastianBoe, line 95:
```suggestion
    if (counter_value >= (uint16_t) img_security_cnt){
```

Might as well skip some execution for this case.

boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
+/*
+ * Copyright (c) 2022 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-BSD-5-Clause-Nordic
+ */
+#include <stdint.h>
```
SebastianBoe, line 6:
```suggestion
/** @file
 * @brief This source file is an adapter layer from MCUBoot's boot_nv_security_counter API onto the NCS/NSIB library bl_otp_counters.
 */

#include <stdint.h>
```

SebastianBoe: might need some line breaks / formatting

boot/zephyr/CMakeLists.txt
```diff
@@ -365,3 +365,9 @@ zephyr_library_sources(
   ${BOOT_DIR}/zephyr/nrf_cleanup.c
 )
 endif()
+
+if(CONFIG_MCUBOOT_HW_DOWNGRADE_PREVENTION)
+zephyr_library_sources(
+  ${BOOT_DIR}/zephyr/nrf_nv_counters.c
+)
+endif()
```
SebastianBoe, line 373: We can avoid some forking if we move nrf_nv_counters.c into nrf and also move this CMake code into nrf.

boot/bootutil/src/loader.c
```diff
@@ -2237,10 +2237,15 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
         }
 #endif /* MCUBOOT_VALIDATE_PRIMARY_SLOT */
 
+#ifdef PM_S1_ADDRESS
+	if (BOOT_CURR_IMG(state) != 1)
```
SebastianBoe, line 2241: Let us document what's going on here to a degree comparable with line 2210.

SebastianBoe (changes requested):
Changes requested

boot/zephyr/nrf_nv_counters.c
Vge0rge: That makes sense, I can do this.

boot/zephyr/nrf_nv_counters.c
Vge0rge: That doesn't work because of the different types.  The security_cnt variable is a fih_int and the counter variable supported by get_monotonic_counter is uint16_t. The problem as I see is that if we cast the variable in the function call the internal function will not know that it is an int variable so it will not pad the rest of the bytes with zeros. I actually tried your approach before and it failed so I will not change this.

boot/zephyr/nrf_nv_counters.c
```diff
@@ -0,0 +1,111 @@
...
+#include "bootutil/fault_injection_hardening.h"
+
+fih_int boot_nv_security_counter_init(void){
+    return FIH_SUCCESS;
+}
+
+/**
+ * Reads the stored value of a given image's security counter.
+ *
+ * @param image_id          Index of the image (from 0).
```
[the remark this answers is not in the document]
Vge0rge: All of the description is part of the upstream header file. It was my mistake to not remove them from the source file but I will do it now. Sorry for the inconvenience.

boot/zephyr/nrf_nv_counters.c
Vge0rge: Will be removed, as I said before

boot/zephyr/nrf_nv_counters.c
Vge0rge: Will be removed, as I said before

boot/zephyr/nrf_nv_counters.c
Vge0rge: Will be removed, as said before.

boot/zephyr/nrf_nv_counters.c
Vge0rge: Will be removed, as said before.

boot/zephyr/nrf_nv_counters.c
Vge0rge: I will do this, yes

boot/zephyr/nrf_nv_counters.c
Vge0rge: Yep

boot/zephyr/pm.yml
Vge0rge: The reason that we need this is that MCUBOOT counters can exist without bl_storage. If we don't use the NSIB but we still want to use the hardware counters for MCUBOOT the partition will not be generated since the SECURE_BOOT_STORAGE is not generated.

boot/zephyr/nrf_nv_counters.c
Vge0rge: Sure

boot/zephyr/CMakeLists.txt
SebastianBoe: and also reduce the amount of manifest PRs we need to deal with.

SebastianBoe (changes requested):
The commit message doesn't match the first hunk (line 2280).

SebastianBoe (approved):

boot/bootutil/src/loader.c
```diff
@@ -2270,11 +2270,11 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
...
 #ifdef PM_S1_ADDRESS
 	/* Patch needed for NCS. Image 1 primary is the currently
-	 * executing MCUBoot image, and is therefore already trusted and
-	 * does not need validation.
+	 * executing MCUBoot image, and is therefore already validated by NSIB and
+	 * does not need to also be validated by MCUBoot.
 	 */
-    bool image_needs_validation = BOOT_CURR_IMG(state) != 1;
-	if (image_needs_validation)
+    bool image_validated_by_nsib = BOOT_CURR_IMG(state) != 1;
```
joerchan, line 2276: This does not have the right meaning AFAICT.

```suggestion
    bool image_not_validated = BOOT_CURR_IMG(state) != 1;
```

Vge0rge: I think that image_to_validate or something similar is better, having negatives on these variables can be confusing.

boot/bootutil/src/loader.c
```diff
@@ -2270,11 +2270,11 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
...
 	/* Patch needed for NCS. Image 1 primary is the currently
-	 * executing MCUBoot image, and is therefore already trusted and
-	 * does not need validation.
+	 * executing MCUBoot image, and is therefore already validated by NSIB and
+	 * does not need to also be validated by MCUBoot.
 	 */
-    bool image_needs_validation = BOOT_CURR_IMG(state) != 1;
-	if (image_needs_validation)
+    bool image_validated_by_nsib = BOOT_CURR_IMG(state) == 1;
+	if (! image_validated_by_nsib)
```
vili-nordic, line 2277: Space after ! ?

joerchan, line 2277:
```suggestion
	if (!image_validated_by_nsib)
```

boot/bootutil/src/loader.c
```diff
@@ -2270,11 +2270,11 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
...
 #ifdef PM_S1_ADDRESS
 	/* Patch needed for NCS. Image 1 primary is the currently
-	 * executing MCUBoot image, and is therefore already trusted and
-	 * does not need validation.
+	 * executing MCUBoot image, and is therefore already validated by NSIB and
+	 * does not need to also be validated by MCUBoot.
 	 */
-    bool image_needs_validation = BOOT_CURR_IMG(state) != 1;
-	if (image_needs_validation)
+    bool image_validated_by_nsib = BOOT_CURR_IMG(state) == 1;
```
vili-nordic, line 2276: Weird indentation?

joerchan: There is spaces here. Change to Tab

SebastianBoe: I changed to spaces as that was what was used by surrounding code.

image

boot/bootutil/src/loader.c
SebastianBoe: removed space

SebastianBoe: fixed

boot/bootutil/src/loader.c
joerchan: Fair enough.

joerchan (approved):

vili-nordic (approved):

de-nordic (approved):

PR #228, opened 2022-12-12
[nrf noup] bootutil: Fix missing PCD define check

nordicjm:
Fixes a missing PCD define check, an image might have the network core partition layout set but if PCD support is not enabled then it should not assume that PCD support is part of mcuboot.


de-nordic (approved):

nordicjm:
@hakonfam @sigvartmh @nvlsianpu please review

sigvartmh (approved):

PR #231, opened 2022-12-15
[nrf noup] treewide: Fix old Nordic license ID

doki-nordic:
NCS has switch to the new license ID some time ago from LicenseRef-BSD-5-Clause-Nordic to new (more
accurate) LicenseRef-Nordic-5-Clause. All source files must be adjusted to the new name.

Ref: NCSIDB-717


carlescufi:
Should this be a `fromlist` instead?

de-nordic (approved):

de-nordic:
> Should this be a `fromlist` instead?

I think these are only applied to things that have been noup previously. But I may be wrong here.

carlescufi (approved):

PR #235, opened 2023-01-25
[nrf noup] boot: zephyr: let secondary_1 slot be in regular flash

sigvartmh:
Added support for multi-image DFU for nRF53 from internal flash which requires the region parameter of `mcuboot_secondary_1` to be configurable. So that the `mcuboot_secondary_1` partition is put into internal flash.

Ref. NCSDK-12809


sigvartmh:
@hakonfam / @nvlsianpu This one needs a review.

boot/zephyr/pm.yml
```diff
@@ -81,7 +81,12 @@ mcuboot_primary_1:
 
 #if (CONFIG_NRF53_MULTI_IMAGE_UPDATE)
 mcuboot_secondary_1:
+#if (CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY_1)
   region: external_flash
+#endif /* CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY_1 */
   size: CONFIG_NRF53_RAM_FLASH_SIZE
+  placement:
```
hakonfam, line 88: We need to make sure that the alignment for internal flash is the same as for mcuboot primary. That should be page aligned. Also we need to make this block different between external flash and internal flash configurations.

hakonfam (dismissed):

boot/zephyr/pm.yml
sigvartmh: Do you know why the flash alignment in external flash is set to 4 by default then?
ref: https://github.com/nrfconnect/sdk-mcuboot/blob/ccc722742bf7d54350f0b8b59c8bead0426e5e56/boot/zephyr/pm.yml#L24

 This could potentially be a misconfiguration and it should be page aligned regardless. 
 Also should I look into potentially using
 ```
 zephyr/dts.cmake:set_target_properties(devicetree_target PROPERTIES "DT_REG|/soc/flash-controller@4001e000|SIZE" "0x1000;")
 ```
 
 For setting the alignment requirements so that it would be page aligned regardless of the device used.

boot/zephyr/pm.yml
```diff
@@ -81,7 +81,12 @@ mcuboot_primary_1:
...
+    after: mcuboot_secondary
```
nvlsianpu, line 89: Isn't this unproper for if `CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY_1` is selected?

sigvartmh: Why? You have both secondary in external flash and placing the secondary_1 after secondary makes sense.

nvlsianpu: The first sentence was obvious. The second one sentence describes something i could anticipate, but didn't know for sure.

nvlsianpu (approved):

de-nordic (approved):

nordicjm (approved):

sigvartmh:
Seems it can never be happy with the file:

```
ERROR   : Test YAMLLint failed: 
YAMLLint (line-length):line too long (110 > 100 characters)
File:boot/zephyr/pm.yml
Line:28
Column:101
ERROR   : Test YAMLLint failed: 
YAMLLint (key-duplicates):duplication of key "mcuboot_secondary" in mapping
```

This will always fail as pm.yml expect to be processed by the preprocessor before fed in as YAML.

PR #239, opened 2023-02-08

sigvartmh:
When mcuboot_secondary is on external flash, the image header cannot dircetly be accessed via secondary_fa->fa_off. Instead the provided function boot_img_hdr() is used now.
Additionally a similar issue is present when trying to read the address of the reset handler. For this flash_area_read() is used now.

With this patch is possible to have the update partiton mcuboot_secondary on external flash and update a updatable bootloader (mcuboot) in s0 and/or s1.

Signed-off-by: Ole Sæther <ole.saether@nordicsemi.no>

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -917,10 +917,9 @@ boot_validated_swap_type(struct boot_loader_state *state,
 #if defined(PM_S1_ADDRESS) || defined(CONFIG_SOC_NRF5340_CPUAPP)
     const struct flash_area *secondary_fa =
         BOOT_IMG_AREA(state, BOOT_SECONDARY_SLOT);
-    struct image_header *hdr = (struct image_header *)secondary_fa->fa_off;
-    uint32_t vtable_addr = 0;
-    uint32_t *vtable = 0;
+    struct image_header *hdr = boot_img_hdr(state, BOOT_SECONDARY_SLOT);
     uint32_t reset_addr = 0;
+    int rc = 0;
     /* Patch needed for NCS. Since image 0 (the app) and image 1 (the other
      * B1 slot S0 or S1) share the same secondary slot, we need to check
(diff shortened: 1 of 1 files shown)


sigvartmh:
@richesonk @chrta FYI.

@nvlsianpu: Resubmitted this patch for NCS 2.3.0 feel free to contact me regarding the matter use-case  as I can't find information on it.

boot/bootutil/src/loader.c
```diff
@@ -930,16 +929,19 @@ boot_validated_swap_type(struct boot_loader_state *state,
      */
 
     if (hdr->ih_magic == IMAGE_MAGIC) {
-        vtable_addr = (uint32_t)hdr + hdr->ih_hdr_size;
-        vtable = (uint32_t *)(vtable_addr);
-        reset_addr = vtable[1];
+        rc = flash_area_read(secondary_fa, hdr->ih_hdr_size +
+				 sizeof(uint32_t), &reset_addr,
+				 sizeof(reset_addr));
```
de-nordic, line 934: MCUboot source expands tabs to four spaces. Why `sizeof(uint32_t)` and then `sizeof(reset_addr)` for the same variable?

sigvartmh: I think this probably was reviewed in the previous PR which was made for this patch.
I think the reasoning is to show the `offset=(hrd_address + hrd_size + reset_address_offset)` of the address of the [reset address](https://documentation-service.arm.com/static/5ea823e69931941038df1b02?token=)  in the ARM vector table stored in the start of a firmware image on cortex-m processors. While the other shows the read-size of the variable.

I'll expand the tabs.

de-nordic (approved):

hakonfam (approved):

nordicjm (approved):

nvlsianpu (approved):

PR #242, opened 2023-02-20
NCS v2.3 rebase

mbolivar-nordic:
Regular history rewrite


tmon-nordic (commented):
While this squashes few fixups and other commits not explicitly marked as fixups, the base commit is unchanged since last history rewrite. Not sure how much value is there in the sequashes alone.

mbolivar-nordic:
> Not sure how much value is there in the sequashes alone.

We can debate this, but our policy is to do the squashes.

plskeggs (approved):

PR #244, opened 2023-03-06
End of history notification

de-nordic:
This is final ~~tag~~ commit in this history line for MCUboot fork, nrfconnect/sdk-mcuboot. The tagged commit, and the old history will be accessible via this tag and on branch v1.9.99-ncs4-snapshot.

The reason of reset of history is described upstream, as Zephyr issue https://github.com/zephyrproject-rtos/zephyr/issues/55303

Briefly: while doing some synchronizations we have brought commits that we did not want in Zephyr fork and later synchronization has duplicated some commits in history; the commits we did not wanted have been reverted upstream anyway and bringing reverts of something we have reverted is again problematic.

That is why we have decided to reset the history by overwriting main branch of the nrfconnect/sdk-mcuboot with zephyrproject-rtos/mcuboot main branch, with the 59624334748129cb93f096408911a227b0dd64c0 top commit.
The nRF specific changes, in form of [nrf noup] commits, will be ported from the v1.9.99-ncs4-snapshot branch on top of the new history.

**Update** Pushed change has corrected tag->commit in commit message, as visible above.
**Update 2023-03-14** After discussion with Marti I have decided to drop information on branch and only leave info about tag. Tag hes been renamed `snapshot2`, as `snashot1` is already used and using `snapshot` looks out of order.


nordicjm (approved):

carlescufi (approved):

de-nordic:
I am replacing this PR with https://github.com/nrfconnect/sdk-mcuboot/pull/245, as it seems that the sdk-nrf/mcuboot has not yet been broken and there is no need to reset history.

PR #248, opened 2023-05-03
[nrf fromtree] bootutil: Fix FIH return type for EC256

Vge0rge:
The commit 5047f03 breaks the FIH with EC256. This brings the fix from upstream.


olivier-le-sage (approved):

frkv (approved):
LGTM

joerchan (approved):

PR #257, opened 2023-05-22
bootutil: loader.c: Fix network core update external flash with 1 image

sigvartmh:
Using sequential updates on the nRF53 with external flash would cause network core updates to fail. This is because the update would be located in the external flash which is network core cannot access. This fixes the issue by copying the data from the external flash into RAM of the application core which can be read by the network core.

This commit also contains some minor renaming to align with the naming scheme used in the `loader.c` file for flash areas.

Ref: NCSDK-21379

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -52,6 +52,10 @@
 #include <dfu/pcd.h>
 #endif
 
+#if defined(CONFIG_FLASH_SIMULATOR)
+#include <zephyr/drivers/flash/flash_simulator.h>
+#endif
+
 #ifdef MCUBOOT_ENC_IMAGES
 #include "bootutil/enc_key.h"
 #endif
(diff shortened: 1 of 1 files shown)


osaether (approved):
I tried it locally and the PR fixes the issue.

de-nordic:
@sigvartmh You want this in ncs 2.4?

hakonfam (approved):

de-nordic:
@sigvartmh @hakonfam Is this targeting NCS 2.4?

sigvartmh:
Yes. I wasn't sure if I should add the milestone or not. Added it now.

de-nordic (approved):

de-nordic:
> @de-nordic Yes. I wanted to know if I should add the milestone or not. Just so you know – I added it now.

You can always add the milestone, if you consider fix to be needed. The milestone can be removed by people responsible for release.

nvlsianpu (approved):

PR #258, opened 2023-05-24
[nrf noup] boot: zephyr: fix serial update for Thingy:91

maxd-nordic:
This patch sets a KConfig value whose default changed to an incompatible value resulting serial update to fail when using the Programmer app in nRF Connect for Desktop.


jhn-nordic (approved):

nordicjm (changes requested):
This should be fixed upstream if it is an upstream issue e.g. because of the use of CDC ACM, but really it doesn't look like an issue in mcuboot at all, the maximum size of a valid MCUmgr packet is 127 bytes, this is mandated by the protocol, if a larger packet is sent then it should rightfully be dropped because it's invalid. So if there is an nordic application which does not respect this, it needs fixing

nordicjm:
> > This should be fixed upstream if it is an upstream issue e.g. because of the use of CDC ACM, but really it doesn't look like an issue in mcuboot at all, the maximum size of a valid MCUmgr packet is 127 bytes, this is mandated by the protocol, if a larger packet is sent then it should rightfully be dropped because it's invalid. So if there is an nordic application which does not respect this, it needs fixing
> 
> Could it be multiple packets sent in a batch? Would that still be invalid behavior from the Programmer app? The Programmer app also doesn't reset the device after serial update, while we're at it.

Sending batch commands is unsupported, what it could be is that it has a large packet which has been fragmented and it's sending them all in one go, which is not invalid operation but would likely differ to how that data would be received over UART (where each newline would be a different receive) but would need to see a capture (e.g. wireshark) to determine what the actual issue is. Reset should be working, I tested it 1-2 months ago with a thingy53.

jhn-nordic:
Is a tooling issue that will be fixed in the tools.

PR #260, opened 2023-06-02
[nrf noup] booutil: loader: Add support for NSIB and multi-image

sigvartmh:
This adds support for using both NSIB and the multi-image configuration
in MCUboot. Before, this was not possible due to the upgradable bootloader
support through NSIB was using the `UPDATEABLE_IMAGE_NUMBER`
configuration to update the updateable bootloader.

In this commit, we change from using `FLASH_AREA_IMAGE_PRIMARY` to get
the flash area ID to using the bootloader state where we set the flash
area ID of the free updatable bootloader slot if the image is intended
for this slot.

Ref. NCSDK-19223

PR #269, opened 2023-08-10
[nrf noup] Move partition manager definitions to pm_sysflash.h

de-nordic:
Making sysflash.h and pm_sysflash.h more readable.

**Update 2023-08-11 15:53 UTC**
Fixed copyrights.

**Update 2023-08-16 12:14 UTC**
Rebase from ncs/main

### DIFF
--- a/boot/zephyr/include/sysflash/pm_sysflash.h
+++ b/boot/zephyr/include/sysflash/pm_sysflash.h
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2023 Nordic Semiconductor ASA
+ *
+ * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
+ */
+
+#ifndef __PM_SYSFLASH_H__
+#define __PM_SYSFLASH_H__
+/* Blocking the __SYSFLASH_H__ */
+#define __SYSFLASH_H__
+
+#include <pm_config.h>
+#include <mcuboot_config/mcuboot_config.h>
+
+#ifndef CONFIG_SINGLE_APPLICATION_SLOT
+
+#if (MCUBOOT_IMAGE_NUMBER == 1)
+
+#define FLASH_AREA_IMAGE_PRIMARY(x)     PM_MCUBOOT_PRIMARY_ID
+#define FLASH_AREA_IMAGE_SECONDARY(x)   PM_MCUBOOT_SECONDARY_ID
+
+#elif (MCUBOOT_IMAGE_NUMBER == 2)
+
+/* If B0 is present then two bootloaders are present, and we must use
+ * a single secondary slot for both primary slots.
+ */
+#if defined(PM_B0_ADDRESS)
+extern uint32_t _image_1_primary_slot_id[];
+#endif
+#if defined(PM_B0_ADDRESS) && defined(CONFIG_NRF53_MULTI_IMAGE_UPDATE)
+#define FLASH_AREA_IMAGE_PRIMARY(x)             \
+        ((x == 0) ?                             \
+           PM_MCUBOOT_PRIMARY_ID :              \
+         (x == 1) ?                             \
+           PM_MCUBOOT_PRIMARY_1_ID :            \
+           255 )
+
+#define FLASH_AREA_IMAGE_SECONDARY(x)           \
+        ((x == 0) ?                             \
+           PM_MCUBOOT_SECONDARY_ID:             \
+        (x == 1) ?                              \
+           PM_MCUBOOT_SECONDARY_1_ID:           \
+           255 )
+#elif defined(PM_B0_ADDRESS)
+
+#define FLASH_AREA_IMAGE_PRIMARY(x)             \
+        ((x == 0) ?                             \
+           PM_MCUBOOT_PRIMARY_ID :              \
+         (x == 1) ?                             \
+          (uint32_t)_image_1_primary_slot_id :  \
+           255 )
+
+#define FLASH_AREA_IMAGE_SECONDARY(x)           \
+        ((x == 0) ?                             \
+            PM_MCUBOOT_SECONDARY_ID:            \
+        (x == 1) ?                              \
+           PM_MCUBOOT_SECONDARY_ID:             \
+           255 )
+#else
+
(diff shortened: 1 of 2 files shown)


boot/zephyr/include/sysflash/pm_sysflash.h
```diff
@@ -0,0 +1,77 @@
...
+
+#define FLASH_AREA_IMAGE_PRIMARY(x)    PM_MCUBOOT_PRIMARY_ID
+#define FLASH_AREA_IMAGE_SECONDARY(x)  PM_MCUBOOT_SECONDARY_ID
+
+#elif (MCUBOOT_IMAGE_NUMBER == 2)
+
+/* If B0 is present then two bootloaders are present, and we must use
+ * a single secondary slot for both primary slots.
+ */
+#ifdef PM_B0_ADDRESS
```
nordicjm, line 27: This doesn't sound right, b0 should be outside the scope of mcuboot entirely. An (actual customer used) design is b0 and mcuboot on app core, b0 -> mcuboot -> user app, then b0 and app on network core where mcuboot on app core can update network core application, I don't see how such an update would work with the below

de-nordic: > This doesn't sound right, b0 should be outside the scope of mcuboot entirely. An (actual customer used) design is b0 and mcuboot on app core, b0 -> mcuboot -> user app, then b0 and app on network core where mcuboot on app core can update network core application, I don't see how such an update would work with the below

That is how it was before, I have just moved it outside of sysflash.h.
I think there may be more problems requiring addressing then that, we probably should review what we have, regarding the partition manager and MCUboot, and figure what needs fixing.

boot/zephyr/include/sysflash/pm_sysflash.h
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2021 Nordic Semiconductor ASA
```
de-nordic, line 2: Incorrect year

boot/zephyr/include/sysflash/pm_sysflash.h
sigvartmh: This is legacy stuff which was added before MCUBoot supported multi-image updates. This is essentially a hack to have mcuboot move a copy of it self into either s0 or s1 depending on where it was linked against. It gets this information through the linker and the symbol _image_1_primary_slot_id. Where this slot id corresponds to the free s0/s1 slot for mcuboot updates.

I have a PR up to fix it for nRF53 but all in all this should be re-designed but it's not a priority currently AFAIK.
Ref.
https://github.com/nrfconnect/sdk-mcuboot/pull/260

sigvartmh: I don't think we can decouple them completely as b0 depends on MCUBoot doing the flash writing.

nordicjm: b0 doesn't depend on anything, it just boots whatever image is valid with the highest version, do you mean b0n for net core?

sigvartmh: Reverse. MCUBoot depends on having some information about where b0 expects images to be located.

nordicjm: This is a real end product static PM file which runs on an nRF5340-based design, first stage bootloader is b0, second stage bootloader is mcuboot, mcuboot supports updating the user application (on the network core) and the application on the network core https://github.com/LairdCP/zephyr_lib/blob/main/mcuboot_config/pm_static.sb.bl5340_dvk_cpuapp.yml how will this work given this file excludes the network core update slot if b0 is present?

boot/zephyr/include/sysflash/pm_sysflash.h
```diff
@@ -0,0 +1,62 @@
...
+#include <mcuboot_config/mcuboot_config.h>
+
+#ifndef CONFIG_SINGLE_APPLICATION_SLOT
+
+#if (MCUBOOT_IMAGE_NUMBER == 1)
+
+#define FLASH_AREA_IMAGE_PRIMARY(x)    PM_MCUBOOT_PRIMARY_ID
+#define FLASH_AREA_IMAGE_SECONDARY(x)  PM_MCUBOOT_SECONDARY_ID
+
+#elif (MCUBOOT_IMAGE_NUMBER == 2) && defined(PM_B0_ADDRESS) && \
```
nordicjm, line 22: I don't see where the configuration for when multi update is present has gone

boot/zephyr/include/sysflash/pm_sysflash.h
```diff
@@ -0,0 +1,92 @@
...
+#include <mcuboot_config/mcuboot_config.h>
+
+#ifndef CONFIG_SINGLE_APPLICATION_SLOT
+
+#if (MCUBOOT_IMAGE_NUMBER == 1)
+
+#define FLASH_AREA_IMAGE_PRIMARY(x)     PM_MCUBOOT_PRIMARY_ID
+#define FLASH_AREA_IMAGE_SECONDARY(x)   PM_MCUBOOT_SECONDARY_ID
+
+#elif (MCUBOOT_IMAGE_NUMBER == 2)
```
nordicjm, line 22: >= 2 right? Either that or needs an #error for > 2 because of the third image XIP?

de-nordic: I have just move code here, I have not applied any more changes than required for the code to be moved.
This line comes in original form from previously modified sysflash.h

nordicjm (approved):

de-nordic:
@sigvartmh ?

de-nordic:
@nvlsianpu Please take a look.

sigvartmh (approved):
As far as I see, this code is 1:1 with what was present in sysflash so should be fine

boot/zephyr/include/sysflash/pm_sysflash.h
sigvartmh: I guess this needs to be taken up in a separate issue. I would create a task for it if this seems needed.

nvlsianpu (approved):

PR #276, opened 2023-09-22
[noup] booutil: loader: Do not check reset vector for XIP image

de-nordic:
The XIP image, 2, does not have reset vector.


boot/bootutil/src/loader.c
```diff
@@ -816,6 +816,12 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
      * overwriting an application written to the incorrect slot.
      * This feature is only supported by ARM platforms.
      */
+#if MCUBOOT_IMAGE_NUMBER == 3
```
nordicjm, line 819: >= 3

nordicjm (approved):

hakonfam (approved):

PR #277, opened 2023-09-28
loader: Add firmware info version check to downgrade prevention

sigvartmh:
For nRF53, the only existing version number metadata is stored in the `firmware_info` structure in the network core. This utilizes PCD to read out the version number and compares it against the version number found in the secondary slot for the network core.

Ref. NCSDK-21379

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -50,6 +50,10 @@
 
 #if defined(CONFIG_SOC_NRF5340_CPUAPP) && defined(PM_CPUNET_B0N_ADDRESS)
 #include <dfu/pcd.h>
+#ifdef CONFIG_PCD_READ_NETCORE_APP_VERSION
+#include <fw_info_bare.h>
+int pcd_version_cmp_net(const struct flash_area *fap, struct image_header *hdr);
+#endif
 #endif
 
 #ifdef MCUBOOT_ENC_IMAGES
@@ -781,9 +785,21 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
 #if defined(MCUBOOT_OVERWRITE_ONLY) && defined(MCUBOOT_DOWNGRADE_PREVENTION)
     if (slot != BOOT_PRIMARY_SLOT) {
         /* Check if version of secondary slot is sufficient */
-        rc = boot_version_cmp(
-                &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver,
-                &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver);
+
+#if defined(CONFIG_SOC_NRF5340_CPUAPP) && defined(CONFIG_NRF53_MULTI_IMAGE_UPDATE) \
+    && defined(CONFIG_PCD_APP) && defined(CONFIG_PCD_READ_NETCORE_APP_VERSION)
+        if (BOOT_CURR_IMG(state) == 1) {
+            rc = pcd_version_cmp_net(fap, boot_img_hdr(state, BOOT_SECONDARY_SLOT));
+        } else {
+             rc = boot_version_cmp(
+                                 &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver,
+                                 &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver);
+        }
+#else
+	rc = boot_version_cmp(
+			&boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver,
+			&boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver);
+#endif
         if (rc < 0 && boot_check_header_erased(state, BOOT_PRIMARY_SLOT)) {
             BOOT_LOG_ERR("insufficient version in secondary slot");
             flash_area_erase(fap, 0, flash_area_get_size(fap));


sigvartmh:
Not finalized I'm pulling things out into a single function something like boot_cmp_net to hide behind a #define flag .

I understand that this is not the best solution but it's what I've come up with as a stop gap to get something in before RC1.

sigvartmh:
@nordicjm So I've fetched out the idea a bit more, but I'm not 100% happy with the end result. ideally, I would have boot_cmp as a hook with a Regular and unique case.

The problem is that boot_cmp does not take in any state or slot, so you don't know which slot you are doing the comparison against. 

Anyways please review the idea if possible, some one is waiting for this to go in for RC1

boot/bootutil/src/loader.c
```diff
@@ -781,9 +782,16 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
...
+
+	/* Prototype to be moved in to hook */
+	#if defined(CONFIG_SOC_NRF5340_CPUAPP)
+	 rc = boot_version_cmp_net(BOOT_CURR_IMG(state), boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver,
+			           &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver);
+	#else
+	rc = boot_version_cmp(
+			&boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver,
+			&boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver);
+	#endif
```
nordicjm, line 794: This doesn't look right, this will mess up someone marking the secondary slot of the primary image for test/confirmation because for some reason it will check the network core version, despite not having anything to do with the network core? And only applies if MCUboot is in upgrade mode?

boot/bootutil/src/loader.c
```diff
@@ -781,9 +782,16 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
...
     if (slot != BOOT_PRIMARY_SLOT) {
         /* Check if version of secondary slot is sufficient */
-        rc = boot_version_cmp(
-                &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver,
-                &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver);
+
+	/* Prototype to be moved in to hook */
+	#if defined(CONFIG_SOC_NRF5340_CPUAPP)
+	 rc = boot_version_cmp_net(BOOT_CURR_IMG(state), boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver,
+			           &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver);
```
de-nordic, line 789: There should be two checks probably one for net core image and other for not-net cores. This will not look nice but it is the best we can do now probably.
So if image == 1 then net core check, the normal check in all the other cases.

sigvartmh:
To answer both @de-nordic and @nordicjm, the idea here is to do something like this: 

`int boot_read_image_header_hook(int img_index, int slot, struct image_header *img_head)`
So inside, you would either do regular compare or do the boot_cmp_net when `if image == 1` I'll refactor this to reflect this better.

sigvartmh:
@de-nordic @nordicjm So I've reworked it a bit but I'm uncertain if I should push even more of the logic over into the nrf53_hook file.

boot/bootutil/src/loader.c
```diff
@@ -781,9 +782,21 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
 #if defined(MCUBOOT_OVERWRITE_ONLY) && defined(MCUBOOT_DOWNGRADE_PREVENTION)
     if (slot != BOOT_PRIMARY_SLOT) {
         /* Check if version of secondary slot is sufficient */
-        rc = boot_version_cmp(
-                &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver,
-                &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver);
+
+	/* Prototype to be moved in to hook */
+	#if defined(CONFIG_SOC_NRF5340_CPUAPP)
+	if (BOOT_CUUR_IMG(state) == 1) {
```
nordicjm, line 788: Have you actually tried this?

nordicjm (changes requested):

boot/bootutil/src/loader.c
sigvartmh: Yeah reverted a change but added typo sorry.

boot/bootutil/src/loader.c
```diff
@@ -781,9 +783,20 @@ boot_validate_slot(struct boot_loader_state *state, int slot,
 #if defined(MCUBOOT_OVERWRITE_ONLY) && defined(MCUBOOT_DOWNGRADE_PREVENTION)
     if (slot != BOOT_PRIMARY_SLOT) {
         /* Check if version of secondary slot is sufficient */
-        rc = boot_version_cmp(
-                &boot_img_hdr(state, BOOT_SECONDARY_SLOT)->ih_ver,
-                &boot_img_hdr(state, BOOT_PRIMARY_SLOT)->ih_ver);
+
+	#if defined(CONFIG_SOC_NRF5340_CPUAPP)
```
nordicjm, line 787: align the # lines to the left and this also needs to check for PCD support, just because an nrf5340 has 2 updateable images doesn't mean the second is for the app core

nordicjm (changes requested):

boot/bootutil/src/loader.c
sigvartmh: Sure I'll check if NRF53_MULTI_IMAGE and PCD is enabled.

I don't really know how this can be solved more generally as we don't have any extra metadata in the images that indicates that this is for network core currently.

So this is how Successful/Unsuccessful updates for network core will look.
image

sigvartmh: Fixed alignments. So what will happen if the customer uploads app to slot 1 and network core to slot 0.

I'll do the same trick as the other functions are doing:
using the reset address and compare it against the base address of CPUNET

nordicjm (commented):
There is probably a PCD hook file somewhere that would need a check adding too

boot/bootutil/src/loader.c
```diff
@@ -49,7 +49,9 @@
 #include "bootutil/mcuboot_status.h"
 
 #if defined(CONFIG_SOC_NRF5340_CPUAPP) && defined(PM_CPUNET_B0N_ADDRESS)
+#include <fw_info_bare.h>
 #include <dfu/pcd.h>
+int pcd_version_cmp_net(const struct flash_area *fap, struct image_header *hdr);
```
de-nordic, line 54: Question: wouldn't it be more useful if the function would rather extract version out of PCD and comparison would happen on the MCUboot side?

sigvartmh: Doesn't use the same versioning scheme. Netcore uses FW_INFO which is just a single value while MCUBoot does the major.minor.rev semantic versioning.

The end goal is to consolidate them but for now, this is what we can provide. 

The command is running on the MCUBoot side but it fetches the current version number through PCD from the network core and does the comparison on Application(MCUBoot side).

sigvartmh:
So do you belive we can get this in for 2.5.0 @nordicjm @de-nordic ?

nordicjm (approved):

de-nordic:
> So do you belive we can get this in for 2.5.0 @nordicjm @de-nordic ?

Yes, but post rc1.

de-nordic (approved):

PR #279, opened 2023-10-05
[nrf noup] zephyr: Add RAM flash configuration to cache for sysbuild

nordicjm:
Puts the flash simulation configurtion into cache variables that can be used by other applications and CMake code to know specifics on the simulated flash details

### DIFF
--- a/boot/zephyr/CMakeLists.txt
+++ b/boot/zephyr/CMakeLists.txt
@@ -352,3 +352,14 @@ zephyr_library_sources(
   ${BOOT_DIR}/zephyr/nrf_cleanup.c
 )
 endif()
+
+if(SYSBUILD AND CONFIG_PCD_APP)
+    # Sysbuild requires details of the RAM flash device are stored to the cache of MCUboot so
+    # that they can be read when running partition manager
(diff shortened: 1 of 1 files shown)


tejlmand (approved):

boot/zephyr/CMakeLists.txt
```diff
@@ -352,3 +352,14 @@ zephyr_library_sources(
   ${BOOT_DIR}/zephyr/nrf_cleanup.c
 )
 endif()
+
+if(SYSBUILD AND CONFIG_PCD_APP)
+    # Sysbuild requires details of the RAM flash device are stored to the cache of MCUboot so
+    # that they can be read when running partition manager
```
57300, line 358: nit: Tab size and wording
```suggestion
  # Sysbuild requires that details of the RAM flash device are stored to the cache of MCUboot so
  # that they can be read when running partition manager
```

57300 (approved):
Out of curiosity, why can't we reuse similar code from `sdk-nrf`?

https://github.com/nrfconnect/sdk-nrf/blob/bed0015d28f9a25f6a01bd29aca857deeae932ab/modules/mcuboot/CMakeLists.txt#L22-L32

PR #280, opened 2023-10-09
[nrf noup] boards: thingy53: disable GPIO ISR support

nika-nordic:
Change disables GPIO interrupt support in Zephyr GPIO driver, which is not obligatory for MCUboot.
This is needed to reduce memory footprint.


nika-nordic:
CI is red but the failures seem to be unrelated to the changes from this PR

anangl (approved):

kapi-no (approved):

kapi-no:
@de-nordic or @nordicjm can you take a look to see if the GPIO interrupt mode isn't needed for the MCUboot image in the Thingy:53 configuration?

(MCUboot image for Thingy:53 builds correctly in the Twistter CI)

de-nordic:
I have built the Machine Learning application with this PR and it seems that MCUboot serial recovery works ok, and can be triggered while holding the Button1 while resetting device.

nordicjm (approved):

nordicjm:
MCUboot does not use GPIO interrupts, it just monitors the pin at startup, would be good to have this enabled when compiling MCUboot for any nordic device...

de-nordic (approved):

PR #284, opened 2023-10-16
sysflash: pm_sysflash: Fix incorrect include of NSIB image define

sigvartmh:
A mismatch in defines was made for `_image_1_primary_slot_id` resulting in some configurations not working. This fixes that the linker variable is exposed for mcuboot so that it knows which slot is running an which slot a bootloader upgrade is to be put into.

Ref. NCSDK-19223

### DIFF
--- a/boot/zephyr/include/sysflash/pm_sysflash.h
+++ b/boot/zephyr/include/sysflash/pm_sysflash.h
@@ -15,13 +15,15 @@
 
 #ifndef CONFIG_SINGLE_APPLICATION_SLOT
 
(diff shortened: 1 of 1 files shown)


nordicjm (approved):

de-nordic (approved):

de-nordic:
@sigvartmh sdk-nrf manifest update PR needed here.

sigvartmh:
manifest PR here https://github.com/nrfconnect/sdk-nrf/pull/12726

PR #300, opened 2024-02-21
[nrf noup] boot/bootutil/loader: introduced cleanup of unusable secondary slot

nvlsianpu:
…dary slot

Added procedure which clean-up content of all the secondary slot which contains valid header but couldn't be assigned to any of supported primary images.
This behavior is needed when configuration allows to use one secondary slot for collecting image for multiple primary slots.

ref.: NCSIDB-1151

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -951,6 +951,87 @@ done:
 }
 #endif /* MCUBOOT_HW_ROLLBACK_PROT */
 
+#if defined(CONFIG_MCUBOOT_CLEANUP_UNUSABLE_SECONDARY) &&\
+(defined(PM_S1_ADDRESS) || defined(CONFIG_SOC_NRF5340_CPUAPP))
+
+#define SEC_SLOT_VIRGIN 0
+#define SEC_SLOT_TOUCHED 1
+#define SEC_SLOT_ASSIGNED 2
+
+#if (MCUBOOT_IMAGE_NUMBER == 2) && defined(PM_B0_ADDRESS) && \
+      !defined(CONFIG_NRF53_MULTI_IMAGE_UPDATE)
+/* This configuration is peculiar - the one physical secondary slot is
+ * mocking two logical secondary
+ */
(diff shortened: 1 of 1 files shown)


boot/bootutil/src/loader.c
```diff
@@ -950,6 +950,76 @@ boot_update_security_counter(uint8_t image_index, int slot,
 }
 #endif /* MCUBOOT_HW_ROLLBACK_PROT */
 
+#if defined(CONFIG_MCUBOOT_CLEANUP_UNUSABLE_SECONDARY) &&\
+defined(PM_S1_ADDRESS) || defined(CONFIG_SOC_NRF5340_CPUAPP)
```
gchwier, line 954: Do we really need this `#if` section? This issue can be reproduced also on other dks (e.g. nrf52) when upload image without b0.

nvlsianpu: -> when upload image built for configuration without b0.

good point

nvlsianpu: @gchwier This code is reachable for nRF52 as well if it supports PM_S1_ADDRESS.

michalek-no (approved):

sigvartmh (approved):

boot/bootutil/src/loader.c
```diff
@@ -951,6 +951,87 @@ boot_update_security_counter(uint8_t image_index, int slot,
...
+{
+}
+static inline void sec_slot_mark_assigned(struct boot_loader_state *state)
+{
+}
+static inline void sec_slot_cleanup_if_unusable(void)
+{
+}
+#endif /* defined(CONFIG_MCUBOOT_CLEANUP_UNUSABLE_SECONDARY) &&\
+          defined(PM_S1_ADDRESS) || defined(CONFIG_SOC_NRF5340_CPUAPP) */
```
gchwier, line 1033: nit: comment should be also changed after updating preprocessor directive

de-nordic (approved):

PR #313, opened 2024-05-17
[nrf noup] loader: remove cleanup for direct xip mode

gchwier:
Move ifdefs just to not add code for cleanup unusable slot when direct xip mode is enabled to avoid warnings.


gchwier:
fix provided also here #314

nvlsianpu (approved):

nordicjm (approved):

PR #314, opened 2024-05-17
[nrf noup] loader: introduced cleanup of unusable secondary slot

de-nordic:
fixup! [nrf noup] loader: introduced cleanup of unusable secondary slot

sec_slot_cleanup_if_unusable is used only in specific configuration of defines and otherwise triggers warning of unused static function declaration.


nvlsianpu (changes requested):
"cleanup of unusable secondary slot" is unrelated to encryption support

nvlsianpu:
@de-nordic This is obsolete, can be closed I believe.

PR #315, opened 2024-06-03
[nrf noup] zephyr: disabled EXT_API_ATLEAST_OPTIONAL

nvlsianpu:
Disabled at last optional EXT_API when external-crypto is enabled.


nordicjm (commented):
Not following this change, why would a user not want this forwarded to their application? It's optional, if it's not found then it just boots it regardless

de-nordic (approved):
As far as I understand this Kconfig does not make sense in crypto configuration as there are some structures missing anyway.
Ack.

nordicjm (approved):

PR #318, opened 2024-06-14
[nrf noup] boot/../loader: skip downgrade prevention for s1/s0 MCUboot self update

nvlsianpu:
This patch introduces skip on checking downgrade for s1/s0 upgrade image (chain-loaded by NSIB). which is used for upgrade MCUboot instance itself.

Reason is that sdk-mcuboot has not access to semantic version of its own image. I also shouldn't touch HW counter used for hardware downgrade prevention for the application image (which was the case, this was even not right counter). HW counters for s0/s1 image are owned by NSIB because its role is to prevnt dongrades of s0/s1 MCUboot.


nordicjm:
This does not work properly, whilst MCUboot running in s0 can successfully load and apply a s1 update, the s1 MCUboot cannot then load a s0 MCUboot update (which has a higher version)

boot/bootutil/src/loader.c
```diff
@@ -1270,7 +1270,7 @@ static inline void sec_slot_cleanup_if_unusable(void)
 }
 #endif /* defined(CONFIG_MCUBOOT_CLEANUP_UNUSABLE_SECONDARY) &&\
           defined(PM_S1_ADDRESS) || defined(CONFIG_SOC_NRF5340_CPUAPP) */
-
+bool owner_nsib[BOOT_IMAGE_NUMBER];
```
nordicjm, line 1273: should not be a global scope variable, plus define at top of file with other local variables

sigvartmh (approved):
This one skips downgrade prevetions for S0/S1 images and looks reasonable

nordicjm (commented):
This is really not working (using `tfm_psa_template`).
Load normal build - ok.
Load an s1 update with MCUboot package version set to 0.2.0 and fw version set to 2 - ok.
Load an s0 update for the original build which should not work - ok.
Load s0 update for the newer version which should then become default - secure boot invalidates the updates s1 image then boots the OLD s0 image, it's just completely downgraded?!

nvlsianpu:
Overwrite only mode: https://github.com/nrfconnect/sdk-mcuboot/pull/318/files#diff-a004fdf5907914a360ce1e7da94221c0165d1c4c8ba7b8c4eaf67bd63af395c1R1047-R1068
This is puzzle for me. For the fist glance this code should be skipped for s1/s0 update - despite it is not skipped, update is working.

nvlsianpu:
@sigvartmh Can you look once more?

boot/bootutil/src/loader.c
```diff
@@ -2488,6 +2521,11 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
                 rc = boot_perform_update(state, &bs);
             }
             assert(rc == 0);
+            if (is_owner_nsib(state)) {
+#ifdef __ZEPHYR__
```
sigvartmh, line 2525: Since this is a downstream patch is this strictly needed? or could we replace it with NVIC_SystemReset() if we are worried that someone is building our downstream version of MCUBoot with another RTOS?

sigvartmh (approved):
in general I like this solution, it was an improvement on the first suggestion and implemented something I was thinking of after I gave my initial approval that we should probably utilize boot_state.

boot/bootutil/src/loader.c
nordicjm: `sys_reboot` is fine, I don't think anyone would be using this for non-zephyr builds because this is part of ncs specifically and we don't support other rtos's with this code or any of the ncs libraries

nordicjm: Though the one thing I would want checked with this before having this merged is that this is actually ok for customers, because I have no idea what they are using b0 + mcuboot updates for, if it is just for mcuboot bug fixes then this is fine, if it is for being able to change the mcuboot key then this will break booting any applications unless they have serial recovery enabled - but I don't know who would have such information. Sales? FAEs?

sigvartmh: > `sys_reboot` is fine, I don't think anyone would be using this for non-zephyr builds because this is part of ncs specifically and we don't support other rtos's with this code or any of the ncs libraries

Then we could drop the `#ifdef`.

> If they are expecting the old image to boot the first time after updating mcuboot so they can then load a new application update (this is a very dodgy thing because a single power failure or reboot before the new image is loaded will brick the device) then there might need to be a Kconfig to control this

This is a good point so having a KConfig toogle makes sense as maybe some customers who gets their keys compromised would actually retain this feature of not rebooting.

nvlsianpu: I don't know about such a risky cascade update method. I can imagine that rather mcuboot would be update  to a (non existing jet) version with key revocation feature.

nordicjm (approved):

nvlsianpu:
@sigvartmh That was my initial idea (utilize the boot_state). And I think technically better than my patch. I gave up with that just because thought that patch will be too tightly coupled to mcuboot core code.

PR #323, opened 2024-07-24
PSA implementation of ed25519 image verification and x25519 random key encryption/decryption

de-nordic:
The PSA encryption is quite complete.
The mbedTLS configuration is rather dodgy.
There are few requirements from upstream and mcuboot upmerge would be nice.

The TODO for reviewers:
 - check licenses on [noups]

PR #326, opened 2024-08-09
[nrf noup] bootutil: Introduce MCUBOOT_ENC_RAW_PRIVATE_KEY

de-nordic:
When enabled the embedded private key is treated as a raw key,
without ASN1 encoding, which saves on flash by removing the
decoding procedures.

Signed-off-by: Dominik Ermel <dominik.ermel@nordicsemi.no>

This is test of proposal. This has been tested on nrf54l15, with PSA x25519, and depending on options the savings were 296-308 bytes.
There are still needles leftovers as the key is stored in non raw (ASN1) form, just the code is skipping decoding and directly returns index of key within encoded stream. There is also size of encoded key stored, so some additional bytes can be dropped.

The PR is on top of https://github.com/nrfconnect/sdk-mcuboot/pull/323


nvlsianpu (approved):

de-nordic:
This has already upstream work done in form of ASN1 bypass.

PR #333, opened 2024-09-09
WIP: mcuboot: Autoconfig fix for pm for MCUboot

frkv:
-Changing include from autoconf.h => zephyr/autoconf.h because the
 former is deprecated in upstream. WIP because this doesn't seem
 like it should be a [nrf noup]


nordicjm:
Is https://github.com/mcu-tools/mcuboot/pull/1987 needed too?

PR #352, opened 2024-10-17
[nrf noup] Remove setting default MCUboot mbedTLS config

de-nordic:
This has been overriding logic that selects nrf-config.h when CONFIG_NRF_SECURITY=y.


nvlsianpu (approved):
The same value is 'the most default one' as set in MCUboot own Kconfig.

nordicjm (approved):

PR #353, opened 2024-10-22
[nrf noup] boot: zephyr: decompression: Fix issues with ARM thumb filter

nordicjm:
fixup! [nrf noup] zephyr: Add support for ARM thumb filter

Fixes some issues with the ARM filter implementation which has been updated to support jump instructions that are staggered into 2 different data chunks, and whereby the final portion of data was not ran through the filter

### DIFF
--- a/boot/zephyr/decompression.c
+++ b/boot/zephyr/decompression.c
@@ -29,6 +29,14 @@
 #define EXPECTED_SIG_TLV IMAGE_TLV_ED25519
 #endif
 
+#define DECOMP_BUF_SIZE CONFIG_BOOT_DECOMPRESSION_BUFFER_SIZE
+#if defined(CONFIG_NRF_COMPRESS_ARM_THUMB)
+#define DECOMP_BUF_EXTRA_SIZE 2
+#else
+#define DECOMP_BUF_EXTRA_SIZE 0
+#endif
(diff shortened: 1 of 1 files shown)


boot/zephyr/decompression.c
```diff
@@ -1044,25 +1064,70 @@ int boot_copy_region_decompress(struct boot_loader_state *state, const struct fl
...
-                    if (rc != 0) {
-                        BOOT_LOG_ERR(
-                            "Flash write failed at offset: 0x%x, size: 0x%x, area: %d, rc: %d",
-                            (off_dst + hdr->ih_hdr_size + write_pos), sizeof(decomp_buf),
-                            fap_dst->fa_id, rc);
-                        rc = BOOT_EFLASH;
-                        goto finish;
-                    }
+                        if ((filter_writeback_pos % sizeof(uint32_t)) != 0)
+                        {
```
de-nordic, line 1086: OK, so the assumption is that filter_writeback_pos is aligned to at least two bytes? because modulo here can give 1 and 3 as well.

nordicjm: The ARM thumb filter takes in data that is 4-byte aligned and outputs data that is 2-byte aligned so 1 and 3 are not possible

de-nordic (approved):
Looks OK.

boot/zephyr/decompression.c
nvlsianpu: ^^ I was thought so.,

PR #363, opened 2024-10-25
Set of PSA encryption fixes

de-nordic:
Fixes in Kconfig, CMake and on in encryption source where one line got lost, somehow, when moving encryption to separate file.

X25519/AES encryption on nrf54l15 PSA has been retested on target and works fine

backport of #362


rlubos:
Doesn't this apply on main too? We should be backporting fixes that were integrated on main first, unless they're really unique to the release branch only.

nvlsianpu:
@rlubos This is backport of #362

nvlsianpu (approved):

gchwier (approved):

nordicjm (approved):

PR #385, opened 2024-12-05
NCS 2.9.0 history cleanup

jukkar:


nordicjm (approved):

fundakol (approved):

jukkar:
There are some checks failing for using fromlist instead of fromtree, but I think I will leave these as is.

PR #394, opened 2025-01-03
scripts: imgtool: fix compression with encryption

michalek-no:
adds TLV which stores compressed image size in case encryption is on. This is to avoid wrong streaam size due to encryption padding.

### DIFF
--- a/boot/bootutil/include/bootutil/image.h
+++ b/boot/bootutil/include/bootutil/image.h
@@ -124,6 +124,7 @@ struct flash_area;
                                             * the format and size of the raw slot (compressed)
                                             * signature
                                             */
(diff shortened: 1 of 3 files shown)


de-nordic (approved):

nvlsianpu (approved):

nvlsianpu:
@michalek-no Are you going to fix git history issue signaled by the `Commit tag`?

michalek-no:
> @michalek-no Are you going to fix git history issue signaled by the `Commit tag`?

This probably will be closed after upmerge as not needed. Otherwise yes.

michalek-no:
already in

PR #396, opened 2025-01-24
[nrf nuop] boot: zephyr: boards: Disabled NCS boot banner for thingy 53

kkasperczyk-no:
Disabled NCS BOOT BANNER to save some flash, as Thingy:53 stopped to fit in the mcuboot partition. The boot banner is not used anyway, as logs are disabled.


maciejbaczmanski:
pulled into https://github.com/nrfconnect/sdk-mcuboot/pull/393 , Thanks!

PR #398, opened 2025-02-07
[nrf fromlist] scripts: imgtool: pretended encryption flag

michalek-no:
allows to set encryption flag without applying encryption.


michalek-no:
duplicate of --clear

PR #402, opened 2025-03-07
[nrf noup] boot: bootutil: loader: Fix monotomic counter update issues

nordicjm:
nrf-squash! [nrf noup] boot/../loader: skip downgrade prevention for s1/s0

Fixes 2 issues with monotomic counter usage:
1. Where the NSIB update skipped the check but would then wrongly update the monotomic counter after
2. Where a network core update on nRF5340 used the monotomic counter which only supports a single image


boot/bootutil/src/loader.c
```diff
@@ -1258,6 +1258,20 @@ boot_update_security_counter(uint8_t image_index, int slot,
...
+#if defined(PM_S1_ADDRESS)
+    if (owner_nsib[image_index]) {
+        /* Downgrade prevention on S0/S1 image is managed by NSIB */
+        return 0;
+    }
+#endif
+
+#if defined(CONFIG_SOC_NRF5340_CPUAPP) && CONFIG_MCUBOOT_NETWORK_CORE_IMAGE_NUMBER != -1
+    if (image_index == CONFIG_MCUBOOT_NETWORK_CORE_IMAGE_NUMBER) {
+        /* Downgrade prevention on network core image is managed by NSIB */
```
nvlsianpu, line 1270:
```suggestion
        /* Downgrade prevention on network core image is not managed  */
```

boot/bootutil/src/loader.c
```diff
@@ -2493,6 +2507,13 @@ check_downgrade_prevention(struct boot_loader_state *state)
     }
 #endif
 
+#if defined(CONFIG_SOC_NRF5340_CPUAPP) && CONFIG_MCUBOOT_NETWORK_CORE_IMAGE_NUMBER != -1
+    if (BOOT_CURR_IMG(state) == CONFIG_MCUBOOT_NETWORK_CORE_IMAGE_NUMBER) {
+        /* Downgrade prevention on network core image is managed by NSIB */
```
nvlsianpu, line 2512:
```suggestion
        /* Downgrade prevention on network core image is not managed */
```

boot/bootutil/src/loader.c
nordicjm: It doesn't do hardware checks but it does do a software check, an older version cannot be loaded so the comment is correct but could be updated to say it's done in software, thoughts?

nordicjm: Updated text

nordicjm:
@nvlsianpu can you re-review?

nvlsianpu (approved):

gchwier (approved):

de-nordic (approved):

PR #406, opened 2025-03-19
[nrf noup] zephyr: Allow hash only for nrf54l devices

de-nordic:
Allow hash only, instead of signature, for nrf54l when using PSA crypto.


nordicjm (approved):
can this be squashed with a previous change?

boot/zephyr/Kconfig
```diff
@@ -207,8 +207,10 @@ choice BOOT_SIGNATURE_TYPE
 
 config BOOT_SIGNATURE_TYPE_NONE
 	bool "No signature; use only hash check"
-	select BOOT_USE_TINYCRYPT
-	select BOOT_IMG_HASH_ALG_SHA256_ALLOW
+	select BOOT_USE_TINYCRYPT if !SOC_SERIES_NRF54LX
```
nordicjm, line 210: if this can be combined with 9644dc258228d71e7eff85025e9d67dbc56df789 then should have `nrf-squash! [nrf noup] PSA configuration required changes` in commit message

de-nordic: Good point. I will have to wait for the PSA PR to get in, because I already modify the same commit there, so SHA will change here.

michalek-no (approved):

nvlsianpu (approved):

PR #407, opened 2025-03-25
[nrf noup] ci: Pin GH actions to SHAs

jangalda-nsc:
Jira: NCSDK-32440


karhama (approved):

thst-nordic:
should we open this change in https://github.com/zephyrproject-rtos/mcuboot first?  
and switch this to fromlist (if it is allowed still)?

nordicjm (changes requested):
not convinced on this, the github ones are trusted as is the zephyr one and this is vastly less readable than the tags (note at thomas said above, you would need to submit this upstream and get David's approval)

PR #415, opened 2025-04-09
[nrf fromlist] zephyr: Disable SPI_NOR by default for nrf7002dk/nrf5340/c…

de-nordic:
…puapp

The change allows to build MCUboot for nrf7002dk/nrf5340/cpuapp with default configuration.

### DIFF
--- a/boot/zephyr/boards/nrf7002dk_nrf5340_cpuapp.conf
+++ b/boot/zephyr/boards/nrf7002dk_nrf5340_cpuapp.conf
@@ -1,0 +1,5 @@
+# By default no external SPI device is used for image storage,
+# so SPI-NOR is not needed. This reduces size of MCUboot, remember
+# though that when external image is needed the CONFIG_SPI_NOR
+# has to be enabled.
+CONFIG_SPI_NOR=n


boot/zephyr/boards/nrf7002dk_nrf5340_cpuapp.conf
```diff
@@ -0,0 +1,5 @@
+# By default no external SPI device is used for image storage,
+# so SPI is not needed. This reduces size of MCUboot, remember
```
hellesvik-nordic, line 2:
```suggestion
# so SPI-NOR is not needed. This reduces size of MCUboot, remember
``

File: boot/zephyr/boards/nrf7002dk_nrf5340_cpuapp.conf [hunk 2]
```diff
@@ -0,0 +1,6 @@
+# By default no external SPI device is used for image storage,
+# so SPI-NOR is not needed. This reduces size of MCUboot, remember
+``
```
nordicjm, line 3:
```suggestion
```
and this should be upstreamed and brought in as fromlist/fromtree

nordicjm (changes requested):

nordicjm:
Actually this doesn't really make sense because I was specifically asked to add support for SPI flash programming for nrf7002dk, let me find who asked and ping

hellesvik-nordic:
But we do the same for other boards as well, no?
For other boards, we disable QSPI-NOR / SPI-NOR drivers is disabled in MCUboot, then it is up to samples or features that use the external flash  to enable it, from how I understand this.
For example: https://github.com/nrfconnect/sdk-mcuboot/blob/main/boot/zephyr/boards/nrf52840dk_nrf52840.conf

nordicjm:
> But we do the same for other boards as well, no? For other boards, we disable QSPI-NOR / SPI-NOR drivers is disabled in MCUboot, then it is up to samples or features that use the external flash to enable it, from how I understand this. For example: https://github.com/nrfconnect/sdk-mcuboot/blob/main/boot/zephyr/boards/nrf52840dk_nrf52840.conf

@LuDuda specifically requested nrfutil QSPI flash programming support for nrf7002dk, which I would assume is where an upgrade image is going to be placed. If it's not then I don't see why support for this was required because nothing else should be writing to it from the PC side, it would be module side interaction

hellesvik-nordic:
> specifically requested nrfutil QSPI flash programming support for nrf7002dk, which I would assume is where an upgrade image is going to be placed.

But that is only **if** external flash is used for DFU. 
Currently users have to add configurations if they want to build **without** external flash, or they get errors.
The intuitive thing, that this PR fixes, is that with this users have to add configurations to build **with** external flash.
For example, this fixes NCSDK-32790, where the nRF7002DK does not for the SMP Server sample out of the box, which shows

boot/zephyr/boards/nrf7002dk_nrf5340_cpuapp.conf
```diff
@@ -0,0 +1,6 @@
+# By default no external SPI device is used for image storage,
```
krish2718, line 1: For 3.0.0 we have added a feature to store nRF70 FW patches in external flash accessible via SPI, this change would break that feature, no? 
+ @ArekBalysNordic

hellesvik-nordic: > We do select FLASH https://github.com/nrfconnect/sdk-nrf/blob/main/subsys/net/lib/nrf70_fw_ext/Kconfig#L36 and may that pulls in SPI/QSPI depending on the board?

No. In fact [CONFIG_SPI_NOR](https://docs.nordicsemi.com/bundle/ncs-latest/page/kconfig/index.html#CONFIG_SPI_NOR) depends on CONFIG_FLASH, so it cannot go the other way. I bet the same is true for QSPI_NOR.

krish2718: ok, for QSPI we do select it explicitly, I can update the fix to choose SPI for nrf7002dk.

hellesvik-nordic: To test, I built the wifi/shell sample on this PR with:
>  west build -p -b nrf7002dk/nrf5340/cpuapp -S nrf70-fw-patch-ext-flash -- -DCONFIG_WIFI_NRF70_LOG_LEVEL_DBG=y

And get the following runtime logs, indicating that this PR does not impact fw patch loading. 
PS: This is not testing fw patch DFU

```
[00:00:00.310,058] <inf> spi_nor: mx25r6435f@0: 8 MiBy flash
[00:00:00.322,998] <dbg> wifi_nrf: nrf_wifi_fw_load: Flash area opened with size: 131072, offset: 0
[00:00:00.323,272] <dbg> wifi_nrf: zep_shim_pr_dbg: Signature: 0xdead1eaf
[00:00:00.323,333] <dbg> wifi_nrf: zep_shim_pr_dbg: num_images: 4
[00:00:00.323,394] <dbg> wifi_nrf: zep_shim_pr_dbg: version: 0x1020e00
[00:00:00.323,455] <dbg> wifi_nrf: zep_shim_pr_dbg: feature_flags: 1
[00:00:00.324,279] <dbg> wifi_nrf: nrf_wifi_fw_load: Processing image 0, len: 1252, num_chunks: 1
[00:00:00.324,310] <dbg> wifi_nrf: nrf_wifi_fw_load: Processing chunk 0-0, size: 1252
[00:00:00.324,340] <dbg> wifi_nrf: nrf_wifi_read_and_download_chunk: Reading chunk of size 1252 from offset 60
--- 17 messages dropped ---
[00:00:00.325,347] <dbg> wifi_nrf: nrf_wifi_fw_load: Processing image 1, len: 46784, num_chunks: 6
[00:00:00.325,378] <dbg> wifi_nrf: nrf_wifi_fw_load: Processing chunk 1-0, size: 8192
[00:00:00.325,439] <dbg> wifi_nrf: nrf_wifi_read_and_download_chunk: Reading chunk of size 8192 from offset 1320
[00:00:00.330,505] <dbg> wifi_nrf: nrf_wifi_fw_load: Processing chunk 1-1, size: 8192
[00:00:00.330,535] <dbg> wifi_nrf: nrf_wifi_read_and_download_chunk: Reading chunk of size 8192 from offset 9512
[00:00:00.335,662] <dbg> wifi_nrf: nrf_wifi_fw_load: Processing chunk 1-2, size: 8192
[00:00:00.335,662] <dbg> wifi_nrf: nrf_wifi_read_and_download_chunk: Reading chunk of size 8192 from offset 17704
[00:00:00.340,789] <dbg> wifi_nrf: nrf_wifi_fw_load: Processing chunk 1-3, size: 8192
[00:00:00.340,789] <dbg> wifi_nrf: nrf_wifi_read_and_download_chunk: Reading chunk of size 8192 from offset 25896
...
```

nordicjm: If everyone is happy that this can be disabled by default in MCUboot then PR is ok but needs to be submitted upstream and brought in as fromlist/fromtree

hellesvik-nordic (approved):

nordicjm (approved):

PR #416, opened 2025-04-10
[nrf noup] boot/zephyr/Kconfig: fix MBEDTLS_CFG_FILE value

nvlsianpu:
Zephyr provides "mcuboot-mbedtls-cfg.h" as glue interface for configure mbedts. "config-tls-generic.h" default value was erroneously introduced during a meta codebase synchronization.


nordicjm (approved):
should use backport label next time and it does it automatically

PR #420, opened 2025-04-22
[nrf noup] zephyr: Fix issue with unpadded encrypted updates

nordicjm:
nrf-squash! zephyr: Add support for compressed image updates

Fixes an issue whereby compressed encrypted update images were not padded and the final part of decryption would fail due to not being a length of the block size


gchwier (approved):
downstream test scenario: mcuboot.external_flash.nrf_compress.encryption.ed25519 passed

de-nordic (approved):

PR #424, opened 2025-04-23
[nrf noup] boot: bootutil: key revocation fix

michalek-no:
adds missing valid return state in key revocation procedure.


boot/bootutil/src/loader.c
```diff
@@ -3000,7 +3000,8 @@ context_boot_go(struct boot_loader_state *state, struct boot_rsp *rsp)
     fih_rc = FIH_SUCCESS;
 #if defined(CONFIG_BOOT_KEYS_REVOCATION)
     rc = revoke();
-    if (rc != BOOT_KEY_REVOKE_OK) {
+    if (rc != BOOT_KEY_REVOKE_OK &&
```
nordicjm, line 3003: add to commit description:
```
nrf-squash! [nrf noup] bootutil: key revocation
```

nordicjm (changes requested):

gchwier (approved):
Upgrade works. Must be mention, that to apply old key revocation, another reset is required after confirming upgrade (swap type: none must be executed)

nordicjm (approved):

PR #435, opened 2025-05-14
[nrf noup] bootutil: Add error log on KMU not provisioned

hellesvik-nordic:
An unprovisioned KMU can cause confusion, so improve the logs for this case.

### DIFF
--- a/boot/bootutil/src/ed25519_psa.c
+++ b/boot/bootutil/src/ed25519_psa.c
@@ -126,6 +126,9 @@ int ED25519_verify(const uint8_t *message, size_t message_len,
         }
 
         BOOT_LOG_ERR("ED25519 signature verification failed %d", status);
+        if(status == PSA_ERROR_INVALID_HANDLE) {
+            BOOT_LOG_ERR("This error(-136) could mean that the KMU slot is not provisioned.");
+        }
     }
 
     return ret;


boot/bootutil/src/ed25519_psa.c
```diff
@@ -126,6 +126,9 @@ int ED25519_verify(const uint8_t *message, size_t message_len,
         }
 
         BOOT_LOG_ERR("ED25519 signature verification failed %d", status);
+        if(status == PSA_ERROR_INVALID_HANDLE) {
+          BOOT_LOG_ERR("PSA_ERROR_INVALID_HANDLE(-136) could mean that the KMU slot is not provisioned.");
+        }
```
de-nordic, line 131: Do not do double log, just if/else the thing.

hellesvik-nordic:
@de-nordic, I added CONFIG_MCUBOOT_SETUP_VALIDATION, which **might** be what we talked about. What do you think about this?

nordicjm (changes requested):
4 space indent for C code in mcuboot
tab indent for Kconfigs
No `default n` in Kconfig, that is the default
1x tab followed by 2x spaces for Kconfig help text indent

boot/zephyr/Kconfig
```diff
@@ -1203,4 +1203,12 @@ config MCUBOOT_VERIFY_IMG_ADDRESS
 	  also be useful when BOOT_DIRECT_XIP is enabled, to ensure that the image
 	  linked at the correct address is loaded.
 
+config MCUBOOT_SETUP_VALIDATION
```
de-nordic, line 1206: Would it be possible to name it MCUBOOT_SETUP_VALIDATION?
I am trying to use the same MCUBOOT_SETUP_VALIDATION as a build mode where MCUboot config is only build to check setup and does not build MCUboot capable of booting image.

boot/zephyr/Kconfig
```diff
@@ -1203,4 +1203,12 @@ config MCUBOOT_VERIFY_IMG_ADDRESS
...
+  bool "Add extra checks at boot to validate before booting"
+	depends on LOG 
+  default n
```
nvlsianpu, line 1209: `default n` not needed

boot/zephyr/Kconfig
hellesvik-nordic: Not relevant after all

de-nordic (approved):

boot/bootutil/src/ed25519_psa.c
```diff
@@ -126,6 +126,9 @@ int ED25519_verify(const uint8_t *message, size_t message_len,
         }
 
         BOOT_LOG_ERR("ED25519 signature verification failed %d", status);
+        if(status == PSA_ERROR_INVALID_HANDLE) {
```
nordicjm, line 129: space between if and bracket in c

nordicjm (commented):
my problem with this is that this bloats anyone whose got logging enabled's code on a condition that will only appear on someone's board who doesn't know what they are doing or doesn't read documentation or doesn't use the new (default) kconfig that is being added to automatically provision things

boot/bootutil/src/ed25519_psa.c
hellesvik-nordic: > will only appear on someone's board who doesn't know what they are doing or doesn't read documentation

For these two, I would say that we should still warn for this case, as users who just program a sample can get this error. But then you say: 
> or doesn't use the new (default) kconfig that is being added to automatically provision things 

And with this, Im not sure that this warning is needed after all.
The use-case for this PR is the following: 

"I think that users should be able to enable a bootloader in any project without having to have a full overview over everything that happens. If the default case is that they need to provision manually, then pretty much everyone will get an error first time they try, and we should then tell them that they need to provision, so not everyone has to spend an hour reading docs to find the line saying they should provision."

However, if the devices will be automatically provisioned by default, then I think we can expect users that set CONFIG_MANUAL_PROVISION or whatever should understand the consequence of this, and the use-case for this PR is no longer valid.

@nordicjm Do I understand correctly that from v3.1.0 and onwards, the nRF54L15 KMU will be provisioned when I run "west flash", so that both MCUboot and NSIB will run without any issues first try?

hellesvik-nordic:
A lot less reasons to give another warning for un-provisioned KMU if the KMU is provisioned by default:
https://github.com/nrfconnect/sdk-nrf/pull/22516

gchwier:
> A lot less reasons to give another warning for un-provisioned KMU if the KMU is provisioned by default: [nrfconnect/sdk-nrf#22516](https://github.com/nrfconnect/sdk-nrf/pull/22516)

but it will not be provisioned by default, user must set:
* SB_CONFIG_SECURE_BOOT_GENERATE_DEFAULT_KMU_KEYFILE, for NSIB or
* SB_CONFIG_MCUBOOT_GENERATE_DEFAULT_KMU_KEYFILE for MCUboot,

So still the problem exists.

PR #439, opened 2025-05-22
[nrf noup] boot: zephyr: boards: 54l15 fprotect

michalek-no:
turn protection on by default.


de-nordic:
You want this to happen for DK or SoC ?

nordicjm (changes requested):
`nrf-squash! [nrf noup] boot/zephyr: add nrf54l15dk ext flash configs` needed in git commit body

nvlsianpu (changes requested):
I'm think that enforcing
```
CONFIG_FPROTECT_ALLOW_COMBINED_REGIONS=y
CONFIG_FPROTECT=y
```
will cause issues when building MCUboot along with NSIB on nRF54L15 target.

I think we can enable Mcuboot's `CONFIG_FPROTECT` & `CONFIG_FPROTECT_ALLOW_COMBINED_REGIONS` using a set_config_bool() syntax in nrf/sysbuild/CMakeLists.txt (by checking `SB_CONFIG_SECURE_BOOT`).
It is worth to use also inmmediate Kconfig for settings both as default values for that so it will be possible to modify this by MCUboot Kconfig still.

michalek-no:
rebase

PR #444, opened 2025-06-02
sysflash: Correct ALL_AVAILABLE_SLOTS when MCUBOOT is not image 2

ball-hayden:
Consider an nRF5340 application with the following sysbuild config:

```
SB_CONFIG_BOOTLOADER_MCUBOOT=y
SB_CONFIG_BOOT_SIGNATURE_TYPE_ECDSA_P256=y

SB_CONFIG_PM_EXTERNAL_FLASH_MCUBOOT_SECONDARY=y

SB_CONFIG_SECURE_BOOT_APPCORE=y
SB_CONFIG_SECURE_BOOT_NETCORE=y

SB_CONFIG_NETCORE_APP_UPDATE=y
SB_CONFIG_MCUBOOT_NRF53_MULTI_IMAGE_UPDATE=y
SB_CONFIG_MCUBOOT_MODE_OVERWRITE_ONLY=y
```

In this case, we have a total of 3 updatable images (app, net, mcuboot).

`MCUBOOT_IMAGE_NUMBER = 2` (i.e. `CONFIG_UPDATEABLE_IMAGE_NUMBER = 2`), as the app secondary partition is reused for MCUBoot secondary.

`CONFIG_MCUBOOT_MCUBOOT_IMAGE_NUMBER = 2` as the net core image has index 1.

In this case, we should fall through to the definition of `ALL_AVAILABLE_SLOTS` that caters for 2 updatable images.


de-nordic (changes requested):
Commit message missing [nrf noup] tagging.

de-nordic (approved):

nordicjm (changes requested):
not sure what this is trying to fix but it does not look valid, there are 2 images which the application sees, that is application and network core, for MCUboot itself it has it's own update slot, and that is already handled by this code in the same file:
```
#if CONFIG_MCUBOOT_MCUBOOT_IMAGE_NUMBER != -1
#ifdef CONFIG_NCS_IS_VARIANT_IMAGE
#define MCUBOOT_S0_S1_SLOTS PM_S0_ID, PM_MCUBOOT_SECONDARY_ID,
#else
#define MCUBOOT_S0_S1_SLOTS PM_S1_ID, PM_MCUBOOT_SECONDARY_ID,
#endif
#else
#define MCUBOOT_S0_S1_SLOTS
#endif

...

static inline uint32_t __flash_area_ids_for_slot(int img, int slot)
{
    static const int all_slots[] = {
        ALL_AVAILABLE_SLOTS
        MCUBOOT_S0_S1_SLOTS
    };
    return all_slots[img * 2 + slot];
};
```
therefore the change here does not make sense

ball-hayden:
Thank you @nordicjm.

Please see https://devzone.nordicsemi.com/f/nordic-q-a/113100/nrf5340-mcuboot-issue-with-multi-image-update-when-sb_config_secure_boot_appcore-is-enabled/537541 for reproduction steps for the issue this is trying to solve.

Without the change, `ALL_AVAILABLE_SLOTS` ends up being just the app core image slots and does not include the net core image slots.

nordicjm (changes requested):
Change does not make sense

ball-hayden:
> I cannot replicate this, taking zephyr/samples/subsys/mgmt/mcumgr/smp_svr and configuring for nrf5340dk/nrf5340/cpuapp with -DSB_CONFIG_NETCORE_EMPTY=y -DSB_CONFIG_NETCORE_APP_UPDATE=y -DSB_CONFIG_SECURE_BOOT_APPCORE=y -DSB_CONFIG_SECURE_BOOT_NETCORE=y -DSB_CONFIG_MCUBOOT_NRF53_MULTI_IMAGE_UPDATE=y

Using `NETCORE_EMPTY` isn't really a fair test - the failure case is that the net core image doesn't update, so you wouldn't see this with an empty netcore image.

Please try with `ipc_radio` as the net core image, adding a `VERSION` file so you can see the version number change in the logs.

You should see that the process looks like it succeeds, but if you observe the MCUBoot logs you will see that image 1 (the net core image) did not swap when it should have, and the version printed by the net core will not have changed.

ball-hayden:
> Change does not make sense

Yes. The whole thing does not make sense. I was very confused. I'm still very confused. The indirection is massive, and it took me several days to come to the proposed solution.

ball-hayden:
I'm glad you've been able to see the issue @nvlsianpu.

I still feel like from the application image the images are being incorrectly reported, but I'm happy to accept suggestions for different approaches here.

ball-hayden:
For context, I've used both the nRF Connect app (on iOS and Android) and https://pypi.org/project/smpclient/ while debugging this.

The higher-level MCUMgr Image Group definitely _thinks_ it has set both image 0 and 1 to the test slot, but when MCUBoot is reading the image trailers it finds that image 0 is set to test, but image 1 is not.

nvlsianpu:
We will take action for fixing the bug.

ball-hayden:
I'm afraid Nordic don't seem to be interested @anthony289478 - they maintain that the current behaviour is correct, despite our independent reports of incorrect behaviour and proposed fixes.

@nordicjm please can we have a proper fix if neither of our proposed solutions are valid?

ball-hayden:
Still looking for a solution @nordicjm / @de-nordic please.

ball-hayden:
@nordicjm / @de-nordic your periodic request for hints for a solution that does make sense, please.

This fault continues to be a problem for us.

ball-hayden:
@nordicjm / @de-nordic I'd still like to work towards a solution here - please let me know what I can do?

PR #459, opened 2025-07-01
Fix for  nsib_swap_run where scramble instead of erase has been expected

de-nordic:
Fixing nsib_swap_run using erase instead of scramble

### DIFF
--- a/boot/bootutil/src/swap_nsib.c
+++ b/boot/bootutil/src/swap_nsib.c
@@ -62,7 +62,7 @@ void nsib_swap_run(struct boot_loader_state *state, struct boot_status *bs)
     rc = swap_scramble_trailer_sectors(state, fap_sec);
     assert(rc == 0);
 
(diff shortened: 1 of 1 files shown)


nordicjm (approved):

boot/bootutil/src/swap_nsib.c
```diff
@@ -62,7 +62,7 @@ void nsib_swap_run(struct boot_loader_state *state, struct boot_status *bs)
     rc = swap_scramble_trailer_sectors(state, fap_sec);
     assert(rc == 0);
```
nordicjm, line 63: remove fixup! and put nrf-squash! on new line with the commit that needs to be squashed

nordicjm (changes requested):

gchwier (approved):
tested and works as expected,
no error logs when upgrading S1 image on nrf54l15

nvlsianpu (approved):

nordicjm (approved):

PR #465, opened 2025-07-04
[nrf noup] bootutil: Locking KMU keys

de-nordic:
Lock KMU keys before passing execution to application.

### DIFF
--- a/boot/bootutil/src/ed25519_psa.c
+++ b/boot/bootutil/src/ed25519_psa.c
@@ -165,4 +165,32 @@ out:
     return ret;
 }
 #endif /* CONFIG_BOOT_KMU_KEYS_REVOCATION */
+
+void nrf_crypto_keys_housekeeping(void)
+{
+    psa_status_t status;
+
+    /* We will continue through all keys, even if we have error while
+     * processing any of it. Only doing BOOT_LOG_DBG, as we do not
+     * really want to inform on failures to lock.
+     */
+    for (int i = 0; i < CONFIG_BOOT_SIGNATURE_KMU_SLOTS; ++i) {
+        psa_key_attributes_t attr;
+
+        status = psa_get_key_attributes(kmu_key_ids[i], &attr);
+        BOOT_LOG_DBG("KMU key 0x%x(%d) attr query status == %d",
+                     kmu_key_ids[i], i, status);
+
+        if (status == PSA_SUCCESS) {
+            status = cracen_kmu_block(&attr);
+            BOOT_LOG_DBG("KMU key lock status == %d", status);
+        }
+
+        status = psa_purge_key(kmu_key_ids[i]);
+        BOOT_LOG_DBG("KMU key 0x%x(%d) purge status == %d",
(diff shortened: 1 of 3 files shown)


boot/zephyr/main.c
```diff
@@ -648,6 +648,13 @@ int main(void)
 
     mcuboot_status_change(MCUBOOT_STATUS_BOOTABLE_IMAGE_FOUND);
 
+    /* From this point MCUboot does not need access to crypto keys.
+     * Cleanup backend key objects and apply key access policies that
```
tomi-font, line 652:
```suggestion
     * Clean up backend key objects and apply key access policies that
```

boot/bootutil/src/ed25519_psa.c
```diff
@@ -165,4 +165,32 @@ int exec_revoke(void)
     return ret;
 }
 #endif /* CONFIG_BOOT_KMU_KEYS_REVOCATION */
+
+void nrf_crypto_keys_housekeeping(void)
+{
+    psa_status_t status = PSA_ERROR_BAD_STATE;
```
tomi-font, line 171: Why do we still initialize `status`?

tomi-font (approved):

frkv (approved):
LGTM

boot/bootutil/src/ed25519_psa.c
tomi-font: probably takes some ROM bytes :upside_down_face: but yeah do what you want really just pointing it out

nvlsianpu (approved):

boot/bootutil/src/ed25519_psa.c
```diff
@@ -165,4 +165,32 @@ int exec_revoke(void)
...
+        status = psa_get_key_attributes(kmu_key_ids[i], &attr);
+        BOOT_LOG_DBG("KMU key 0x%x(%d) attr query status == %d",
+                     kmu_key_ids[i], i, status);
+
+        if (status == PSA_SUCCESS) {
+            status = cracen_kmu_block(&attr);
+            BOOT_LOG_DBG("KMU key lock status == %d", status);
+        }
+
+        status = psa_purge_key(kmu_key_ids[i]);
```
frkv, line 189: Commenting here as we have merged an API that is irrelevant for KMU support 
https://arm-software.github.io/psa-api/crypto/1.1/api/keys/management.html#c.psa_purge_key

This type of API is only required when [PSA_KEY_USAGE_CACHE](https://arm-software.github.io/psa-api/crypto/1.1/api/keys/policy.html#c.PSA_KEY_USAGE_CACHE) is in use, which is not the case for NSIB or MCUboot

There is no key caching functionality forwarded to PSA crypto driver wrappers (psa_crypto_driver_wrapper_purge-key doesn't exists). Likewise there is nothing in the thin PSA core used by [bl2](https://github.com/nrfconnect/sdk-trusted-firmware-m/blob/main/bl2/src/thin_psa_crypto_core.c)

PR #467, opened 2025-07-08
boot: Introduce retry mechanism for image verification

adsz-nordic:
Intended mainly for direct-xip mode.
Allows to control:
- number of image validation attempts performed before considering the image invalid
- time before next attempt is made


nvlsianpu (commented):
I'm fine with this feature.

boot/zephyr/Kconfig
```diff
@@ -1213,4 +1213,22 @@ config MCUBOOT_VERIFY_IMG_ADDRESS
 	  also be useful when BOOT_DIRECT_XIP is enabled, to ensure that the image
 	  linked at the correct address is loaded.
 
+
+config MCUBOOT_IMG_VALIDATE_ATTEMPT_COUNT
```
de-nordic, line 1217: Can we standardize to use NRF_ prefix on MCUboot options that are added as NRF noup things?
Here it would be NRF_MCUBOOT_IMG_VALIDATE_ATTEMPT_COUNT. This way we can avoid any collisions with upstream added features.

tomchy (approved):

nordicjm (approved):

PR #471, opened 2025-07-14
zephyr: boards: align nrf54h20 with iron

bjarki-andreasen:
The nrf54h20 iron variant was removed, as nrf54h20 is now iron by default. The overlay for it was not updated though, so that is done with this commit.


de-nordic (approved):

nvlsianpu:
@bjarki-andreasen Is this still valid?

bjarki-andreasen:
> @bjarki-andreasen Is this still valid?
> 

Nope

PR #477, opened 2025-07-24
Zephyr: introduce bootloader requests

tomchy:


boot/bootutil/zephyr/src/boot_request_retention.c
```diff
@@ -0,0 +1,341 @@
...
+#include "bootutil/bootutil_log.h"
+
+/** Special value of image number, indicating a request to the bootloader. */
+#define BOOT_REQUEST_IMG_BOOTLOADER 0xFF
+
+/** Helper value, indicating how many images are supported. */
+#define BOOT_REQUEST_IMAGES 2
+
+/** Number of requests per image. */
+#define BOOT_REQUEST_PER_IMAGE 2
```
nordicjm, line 18: replace with `BOOT_IMAGE_NUMBER` for number of images, `BOOT_NUM_SLOTS` for number of slots, since they already exist

boot/bootutil/zephyr/src/boot_request_retention.c
```diff
@@ -0,0 +1,341 @@
...
+
+    for (size_t i = 0; i < retention_size(bootloader_request_dev); i++) {
+        ret = retention_write(bootloader_request_dev, i * sizeof(value), (void *)&value,
+                              sizeof(value));
+        if (ret != 0) {
+            break;
+        }
+    }
+
+    return ret;
```
nordicjm, line 148:
```suggestion
return retention_clear(bootloader_request_dev);
```

boot/bootutil/zephyr/src/boot_request_retention.c
```diff
@@ -0,0 +1,341 @@
...
+}
+
+bool
+boot_request_detect_firmware_loader(void)
+{
+    uint8_t value = BOOT_REQUEST_MODE_INVALID;
+    size_t req_slot;
+    int ret;
+
+    ret = boot_request_slot_find(BOOT_REQUEST_BOOT_MODE, BOOT_REQUEST_IMG_BOOTLOADER,
```
nordicjm, line 328: #ifdef these with the firmware loader Kconfig or define

boot/zephyr/CMakeLists.txt
```diff
@@ -487,6 +487,10 @@ if(CONFIG_MCUBOOT_BOOT_BANNER)
   zephyr_sources(kernel/banner.c)
 endif()
 
+if(CONFIG_FIND_NEXT_SLOT_HOOKS)
+  zephyr_sources(hooks_find_next_slot.c)
+endif()
```
nordicjm, line 492: need another Kconfig to use this file so that users can keep `CONFIG_FIND_NEXT_SLOT_HOOKS` enabled but disable the other one and replace with their own implementation

nordicjm (changes requested):

boot/bootutil/zephyr/src/boot_request_retention.c
tomchy: Replaced with: `CONFIG_UPDATEABLE_IMAGE_NUMBER` since `BOOT_IMAGE_NUMBER` is not available on the application side.
The `BOOT_REQUEST_PER_IMAGE` is equal to the number of slots only accidentally - it represents the number of different request types that can be stored per image.

boot/bootutil/zephyr/src/boot_request_retention.c
tomchy: Honestly, I do not see much value in adding those `ifdef`s - unreferenced functions will get lost and do not increase the build size and adding those create new inter-module dependencies, that may become painful later.

At this stage I will add them, because we do not have a second use case for the module, so it is hard to judge which option is better.

boot/bootutil/src/bootutil_public.c
```diff
@@ -503,32 +508,93 @@ boot_write_copy_done(const struct flash_area *fap)
     return boot_write_trailer_flag(fap, off, BOOT_FLAG_SET);
 }
 
-#ifndef MCUBOOT_BOOTUTIL_LIB_FOR_DIRECT_XIP
+#ifdef SEND_BOOT_REQUEST
+bool
+boot_send_requests(uint8_t magic, bool active, bool confirm, int image_id, uint32_t slot_id)
```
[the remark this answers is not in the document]
tomchy: Thanks for all of the suggestions!

boot/bootutil/src/bootutil_public.c
```diff
@@ -503,32 +508,93 @@ boot_write_copy_done(const struct flash_area *fap)
...
+        }
+        break;
+
+    case BOOT_MAGIC_BAD:
+    default:
+        rc = BOOT_EBADIMAGE;
+        break;
+    }
+
+    if (active && (rc ==0)) {
```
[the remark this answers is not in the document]
tomchy: I agree with the `active` flag - it is not needed here and using `rc` as a return code is way better.
When it comes to the `magic` part - I am bit worried that confirming an image that does not have a correct magic does not sound right (and - does not follow the order in which flags are written, which may be crucial on flash-based devices), so I prefer to keep it.

boot/bootutil/zephyr/src/boot_request_retention.c
```diff
@@ -0,0 +1,333 @@
...
+        *slot = 1 + image * BOOT_REQUEST_PER_IMAGE;
+        break;
+    case BOOT_REQUEST_IMG_CONFIRM:
+        *slot = 2 + image * BOOT_REQUEST_PER_IMAGE;
+        break;
+    default:
+        return -EINVAL;
+    }
+
+    if (*slot > BOOT_REQUEST_SLOT_MAX_N) {
```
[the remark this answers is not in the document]
tomchy: Nice catch! That's what happens if we write code without unit tests...

boot/bootutil/include/bootutil/boot_request.h
```diff
@@ -0,0 +1,105 @@
...
+ * @return 0 if requested, negative error code otherwise.
+ */
+int boot_request_enter_recovery(void);
+
+/**
+ * @brief Request a bootloader to boot firmware loader image.
+ *
+ * @return 0 if requested, negative error code otherwise.
+ */
+int boot_request_enter_firmware_loader(void);
```
nordicjm, line 52: can both be combined into `boot_request_enter_recovery` there is either serial recovery or firmware loader, not both

boot/bootutil/include/bootutil/boot_request.h
```diff
@@ -0,0 +1,105 @@
...
+ * @return true if requested, false otherwise.
+ */
+bool boot_request_detect_recovery(void);
+
+/**
+ * @brief Check if there is a request to boot firmware loader image.
+ *
+ * @return true if requested, false otherwise.
+ */
+bool boot_request_detect_firmware_loader(void);
```
nordicjm, line 85: as above

boot/bootutil/zephyr/src/boot_request_retention.c
```diff
@@ -0,0 +1,333 @@
...
+/** Maximum number of request slots. */
+#define BOOT_REQUEST_SLOT_MAX_N (BOOT_REQUEST_PER_IMAGE * CONFIG_UPDATEABLE_IMAGE_NUMBER + 1)
+
+MCUBOOT_LOG_MODULE_REGISTER(bootloader_request);
+
+static const struct device *bootloader_request_dev =
+    DEVICE_DT_GET(DT_CHOSEN(zephyr_bootloader_request));
+BUILD_ASSERT((BOOT_REQUEST_SLOT_MAX_N * sizeof(uint8_t)) <
+             DT_REG_SIZE_BY_IDX(DT_CHOSEN(zephyr_bootloader_request), 0),
+             "zephyr,bootloader-request area is too small for bootloader request struct");
```
nordicjm, line 26: change to `nrf,` prefix (same when set in dts)

boot/bootutil/zephyr/src/boot_request_retention.c
```diff
@@ -0,0 +1,333 @@
...
+
+enum boot_request_type {
+    /** Invalid request. */
+    BOOT_REQUEST_INVALID = 0,
```
nordicjm, line 30: nit. enums start at 0 and count up

boot/bootutil/zephyr/src/boot_request_retention.c
```diff
@@ -0,0 +1,333 @@
...
+
+    switch (type) {
+    case BOOT_REQUEST_BOOT_MODE:
+        *slot = 0;
+        break;
+    case BOOT_REQUEST_IMG_PREFERENCE:
+        *slot = 1 + image * BOOT_REQUEST_PER_IMAGE;
+        break;
+    case BOOT_REQUEST_IMG_CONFIRM:
+        *slot = 2 + image * BOOT_REQUEST_PER_IMAGE;
```
nordicjm, line 108: use enum for what these values mean, can actually then have a `_COUNT` in the enum and use that for the number of states instead of just defining to 2 separately

boot/bootutil/zephyr/src/boot_request_retention.c
```diff
@@ -0,0 +1,333 @@
...
+                         sizeof(value));
+    if (ret != 0) {
+        return false;
+    }
+
+    switch (value) {
+    case BOOT_REQUEST_SLOT_PRIMARY:
+        return (slot == 0);
+    case BOOT_REQUEST_SLOT_SECONDARY:
+        return (slot == 1);
```
nordicjm, line 186: there are defines in mcuboot for primary/secondary slot IDs

boot/bootutil/zephyr/src/boot_request_retention.c
```diff
@@ -0,0 +1,333 @@
...
+        return ret;
+    }
+
+    switch (slot) {
+    case 0:
+        value = BOOT_REQUEST_SLOT_PRIMARY;
+        break;
+    case 1:
+        value = BOOT_REQUEST_SLOT_SECONDARY;
+        break;
```
nordicjm, line 155: as above

boot/bootutil/zephyr/src/boot_request_retention.c
```diff
@@ -0,0 +1,333 @@
...
```
nordicjm, line 212: and further below

boot/zephyr/Kconfig
```diff
@@ -1297,6 +1297,10 @@ config USB_DEVICE_PRODUCT
 config MCUBOOT_BOOTUTIL_LIB_OWN_LOG
 	bool
 
+config MCUBOOT_BOOT_REQUEST
```
nordicjm, line 1300: apply Kconfig name changes as per discussion

nvlsianpu, line 1300: config NRF_xxx

boot/zephyr/Kconfig.firmware_loader
```diff
@@ -42,6 +42,12 @@ config BOOT_FIRMWARE_LOADER_PIN_RESET
 	  Checks if the module reset was caused by the reset pin and will
 	  remain in bootloader firmware loader mode if it was.
 
+config BOOT_FIRMWARE_LOADER_BOOT_REQ
```
nvlsianpu, line 45: config NRF_xxx

boot/zephyr/Kconfig.serial_recovery
```diff
@@ -191,6 +191,12 @@ config BOOT_SERIAL_PIN_RESET
 	  Checks if the module reset was caused by the reset pin and will
 	  remain in bootloader serial recovery mode if it was.
 
+config BOOT_SERIAL_BOOT_REQ
```
nvlsianpu, line 194: config NRF_xxx

boot/bootutil/src/bootutil_public.c
nvlsianpu, line 1: just wonder whether flash_area_to_image() -> flash_area_to_image() can more associated with a upstream patch?

boot/bootutil/include/bootutil/boot_request.h
tomchy: I do not really like to enter firmware loader by calling `boot_request_enter_recovery` API.

boot/bootutil/src/bootutil_public.c
tomchy: Yet I do not have an argument, why this change is needed there...

boot/bootutil/zephyr/src/boot_request_retention.c
tomchy: Decided to cover in a separate PR.

boot/bootutil/zephyr/src/boot_request_retention.c
tomchy: Decided to cover in a separate PR.

nordicjm (approved):

PR #493, opened 2025-08-06
Fixed security counter overflow detected to late

ahasztag:
This commit fixes the issue, occuring when the maximum amount of security counter updates has been reached.

This fact was only detected after a permament update already happened - the updated firmware was unable to boot, as it failed when trying to update the security counter after the permament swap.

This commit adds the check if the security counter can be updated (i. e. free security counter slots are still available) before the swap is performed, fixing the issue.

### DIFF
--- a/boot/zephyr/Kconfig
+++ b/boot/zephyr/Kconfig
@@ -1079,6 +1079,7 @@ config MCUBOOT_HW_DOWNGRADE_PREVENTION
 config MCUBOOT_HW_DOWNGRADE_PREVENTION_COUNTER_LIMITED
 	bool "HW based downgrade prevention counter has limited number of updates"
 	depends on MCUBOOT_HW_DOWNGRADE_PREVENTION
+	default y if SOC_NRF5340_CPUAPP || SOC_SERIES_NRF91X || SOC_SERIES_NRF54LX
 	help
 	  When this option is set, the hardware downgrade prevention counter
 	  has limited number of updates. This option will enable checking


ahasztag:
Note: I have not added nrf squash! as there does not seem to be a noup commit this could be reasonably squashed with. The only noup which changes code related to the security counter seems to be https://github.com/nrfconnect/sdk-mcuboot/commit/bf143850d06f24d83c51bc904034e3ccdf22290c but the issue it addresses is unrelated.

Also, I wasn't able to find a way to put this fix upstream. The fix needs a function like the added `boot_nv_security_counter_is_update_possible` which is platform specific and needs to be placed in NCS like the other security counter related platform functions. Adding a call to this function would break any non-NCS platforms. If you think this fix should go upstream please let me know if there is a way to address `boot_nv_security_counter_is_update_possible` being undefined for non-NCS platforms

nordicjm:
> Note: I have not added nrf squash! as there does not seem to be a noup commit this could be reasonably squashed with. The only noup which changes code related to the security counter seems to be [bf14385](https://github.com/nrfconnect/sdk-mcuboot/commit/bf143850d06f24d83c51bc904034e3ccdf22290c) but the issue it addresses is unrelated.
> 
> Also, I wasn't able to find a way to put this fix upstream. The fix needs a function like the added `boot_nv_security_counter_is_update_possible` which is platform specific and needs to be placed in NCS like the other security counter related platform functions. Adding a call to this function would break any non-NCS platforms. If you think this fix should go upstream please let me know if there is a way to address `boot_nv_security_counter_is_update_possible` being undefined for non-NCS platforms

Use a hook

nordic-mik7 (approved):

PR #500, opened 2025-08-07
[nrf noup] zephyr: Remove dependency on key import when KMU is used

de-nordic:
nrf-squash! [nrf noup] bootutil: Add support for KMU stored ED25519 signature key

We are not importing keys, from MCUboot, when KMU is in use.


nvlsianpu (approved):

gchwier (approved):

nvlsianpu (commented):
I think now this can be some more complicated as: PSA encryption (x25519) imports keys.

PR #509, opened 2025-08-08
[nrf fromlist] bootutil: swap_scratch: Fix scratch scrambling

nordic-mik7:
This PR fixes the scrambling of the scratch area after persisting a trailer to the primary slot. Previous use of boot_erase_region() didn't guarantee that the scratch area is scrambled, as it only erases memory for devices that selects CONFIG_MCUBOOT_STORAGE_WITH_ERASE.

Upstream PR #: 2418


shanthanordic:
Removed 3.1.0 milestone, review pending

PR #517, opened 2025-08-22
[nrf noup] Support for ed25519 signature verification using ITS

nordic-mik7:
Thic PR introduces support for ed25519 signature verification when CONFIG_NCS_BOOT_SIGNATURE_USING_ITS is set (through PSA API).

### DIFF
--- a/boot/bootutil/src/ed25519_psa.c
+++ b/boot/bootutil/src/ed25519_psa.c
@@ -26,22 +26,35 @@ BOOT_LOG_MODULE_REGISTER(ed25519_psa);
 #if defined(CONFIG_BOOT_SIGNATURE_USING_KMU)
 /* List of KMU stored key ids available for MCUboot */
 #define MAKE_PSA_KMU_KEY_ID(id) PSA_KEY_HANDLE_FROM_CRACEN_KMU_SLOT(CRACEN_KMU_KEY_USAGE_SCHEME_RAW, id)
-static psa_key_id_t kmu_key_ids[3] =  {
+static psa_key_id_t key_ids[] =  {
     MAKE_PSA_KMU_KEY_ID(226),
     MAKE_PSA_KMU_KEY_ID(228),
     MAKE_PSA_KMU_KEY_ID(230)
 };
+
+#define KEY_SLOTS_COUNT CONFIG_BOOT_SIGNATURE_KMU_SLOTS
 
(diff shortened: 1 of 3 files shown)


ahasztag:
> No. Rename the `kmu_key_ids` to `key_ids` or whatever common name there can be made, and name the both lists the name, just choose one by selected secure storage. Reduce all the changes, in the ifdef from, by moving all required compile time logic into one place, above all the usage cases. Add common cpp identifiers for key count and secure storage usage.

Michał is on vacation so I will continue work on this PR. Changed as you've suggested.

boot/bootutil/src/ed25519_psa.c
```diff
@@ -113,23 +126,24 @@ int ED25519_verify(const uint8_t *message, size_t message_len,
...
                                     message_len, signature,
                                     EDDSA_SIGNAGURE_LENGTH);
         if (status == PSA_SUCCESS) {
             ret = 1;
 #if defined(CONFIG_BOOT_KMU_KEYS_REVOCATION)
-            validated_with = kmu_key_ids + i;
+            validated_with = key_ids + i;
 #endif
-            break;
+            return ret;
```
de-nordic, line 140: Just return 1 here, remove 'ret' and replace final return with 'return 0'.

boot/bootutil/src/ed25519_psa.c
```diff
@@ -26,22 +26,35 @@ BOOT_LOG_MODULE_REGISTER(ed25519_psa);
 #if defined(CONFIG_BOOT_SIGNATURE_USING_KMU)
 /* List of KMU stored key ids available for MCUboot */
 #define MAKE_PSA_KMU_KEY_ID(id) PSA_KEY_HANDLE_FROM_CRACEN_KMU_SLOT(CRACEN_KMU_KEY_USAGE_SCHEME_RAW, id)
-static psa_key_id_t kmu_key_ids[3] =  {
+static psa_key_id_t key_ids[3] =  {
```
de-nordic, line 29: Since the line is already touched, remove the 3.

de-nordic (changes requested):

boot/bootutil/src/ed25519_psa.c
ahasztag: Done

de-nordic (approved):

PR #522, opened 2025-09-01
[nrf noup] boot/zephyrnrf_cleanup: ensure UARTE RX disabled

nvlsianpu:
The workaround in the UARTE driver which was up to disabling RX was removed in zephyr-rtos v4.0. Since that MCUboot needs to do corresponding action.

*note:*
Probably patch might improved a bit so it would be applied only for shell's and mcumgr's console UARTE.

### DIFF
--- a/boot/zephyr/nrf_cleanup.c
+++ b/boot/zephyr/nrf_cleanup.c
@@ -1,8 +1,10 @@
 /*
- * Copyright (c) 2020 Nordic Semiconductor ASA
+ * Copyright (c) 2020-2025 Nordic Semiconductor ASA
  *
  * SPDX-License-Identifier: LicenseRef-Nordic-5-Clause
  */
+
+#include <zephyr/kernel.h>
 
 #if defined(CONFIG_NRFX_CLOCK)
 #include <hal/nrf_clock.h>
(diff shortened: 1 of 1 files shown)


boot/zephyr/nrf_cleanup.c
```diff
@@ -105,6 +107,21 @@ void nrf_cleanup_peripheral(void)
     for (int i = 0; i < sizeof(nrf_uarte_to_clean) / sizeof(nrf_uarte_to_clean[0]); ++i) {
         NRF_UARTE_Type *current = nrf_uarte_to_clean[i];
 
+        /* Ensured that RX is stopped */
+        if (nrfy_uarte_event_check(current, NRF_UARTE_EVENT_RXSTARTED)) {
+            nrfy_uarte_task_trigger(current, NRF_UARTE_TASK_STOPRX);
+            /* Wait up to 100ms for task to stop */
+            for (int i = 0; i < 1000; i++) {
```
nordicjm, line 114: heard yesterday there is an deinit driver function in zephyr from @bjarki-andreasen not sure if it does the same as this or different but might be worth a look at

bjarki-andreasen: The API is indeed implemented https://github.com/zephyrproject-rtos/zephyr/pull/93273 can't we just use it?

PR #526, opened 2025-09-05
[nrf noup] Correct order of #endif comments

de-nordic:
nrf-squash! [nrf noup] Added BOOT_SIGNATURE_USING_ITS for ecdsa configuration

Cosmetic fix.


de-nordic:
No longer needed

PR #531, opened 2025-09-12
[nrf noup] bootutil: Use correct set of KMU key slots

nordicjm:
nrf-squash! [nrf noup] bootutil: Add support for KMU stored ED25519 signature key

Will instead use the immutable bootloader key slot IDs if b0 is not enabled, adds a Kconfig which can be used to fall back to the previous slot IDs for previously deployed bootloaders


boot/bootutil/src/ed25519_psa.c
```diff
@@ -25,11 +25,20 @@ BOOT_LOG_MODULE_REGISTER(ed25519_psa);
 
 #if defined(CONFIG_BOOT_SIGNATURE_USING_KMU)
 /* List of KMU stored key ids available for MCUboot */
+#define PSA_KEY_INDEX_SIZE 2
+
+#if CONFIG_MCUBOOT_MCUBOOT_IMAGE_NUMBER != -1 || \
+	defined(CONFIG_NCS_BOOT_SIGNATURE_KMU_ORIGINAL_MAPPING)
+#define PSA_KEY_STARTING_ID 226
```
de-nordic, line 32: Should we maybe use Kconfig for KMU base address?

nordicjm: would probably make sense and allow for future devices if they have different mappings, but will leave for someone else to implement

eivindj-nordic:
@frkv @nvlsianpu Please review. Would like to get this in to unblock the PR to NCS Bare Metal.

boot/zephyr/Kconfig
```diff
@@ -421,6 +421,17 @@ config BOOT_SIGNATURE_KMU_SLOTS
 	  Selects the number of KMU key slots (also known as generations) to use when verifying
 	  an image.
 
+config NCS_BOOT_SIGNATURE_KMU_ORIGINAL_MAPPING
+	bool "Use original mapping [DEPRECATED]"
+	depends on SOC_SERIES_NRF54LX
+	depends on MCUBOOT_MCUBOOT_IMAGE_NUMBER = -1
+	select DEPRECATED
+	help
+	  When this option is enabled, it will use the previous key slot IDs for the MCUboot image
```
nvlsianpu, line 430:
```suggestion
	  When this option is enabled, it will use the UROT_PUBKEY key slot IDs for the MCUboot image
```

boot/zephyr/Kconfig
```diff
@@ -421,6 +421,17 @@ config BOOT_SIGNATURE_KMU_SLOTS
 	  Selects the number of KMU key slots (also known as generations) to use when verifying
 	  an image.
 
+config NCS_BOOT_SIGNATURE_KMU_ORIGINAL_MAPPING
```
nvlsianpu, line 424: maybe: NCS_BOOT_SIGNATURE_KMU_ORIGINAL_MAPPING -> NCS_BOOT_SIGNATURE_KMU_UROT_MAPPING
?

nvlsianpu (changes requested):
let's be more descriptive,

boot/zephyr/Kconfig
nordicjm: Updated

boot/zephyr/Kconfig
nordicjm: Updated

PR #534, opened 2025-09-16
[nrf noup] boot/zephyr: check for FPROTECT capability

nvlsianpu:
Added run-time check for checking area setup to be protected meets device capability.


michalek-no:
There are few different region sizes across 54L family

PR #541, opened 2025-09-29
[nrf noup] nrf_cleanup: temporary GRTC cleanup with no counter reset

ahasztag:
The GRTC should not reset its value and stop running when switching for
MCUBoot to application. Also in general, this should not be done in case
of reset - so the nrfx_grtc_unit function should not reset the counter
value. This will be fixed in nrfx (NRFX-8487), but the solution
is needed quicker. Hence, a temporary uninit function for GRTC was
added in nrf_cleanup, fixing this issue for MCUBoot.

### DIFF
--- a/boot/zephyr/nrf_cleanup.c
+++ b/boot/zephyr/nrf_cleanup.c
@@ -52,9 +52,35 @@ static inline void nrf_cleanup_rtc(NRF_RTC_Type * rtc_reg)
 #endif
 
 #if defined(CONFIG_NRF_GRTC_TIMER)
+
+/**
+ * This function is temporary and should be removed once nrfx_grtc_uninit
+ * no longer resets the counter - see NRFX-8487.
+ */
+static inline void nrfx_grtc_uninit_no_counter_reset(void)
+{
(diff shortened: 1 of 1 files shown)


boot/zephyr/nrf_cleanup.c
```diff
@@ -51,13 +48,6 @@ static inline void nrf_cleanup_rtc(NRF_RTC_Type * rtc_reg)
 }
 #endif
 
-#if defined(CONFIG_NRF_GRTC_TIMER)
-static inline void nrf_cleanup_grtc(void)
-{
-    nrfx_grtc_uninit();
```
nordicjm, line 57: I don't follow this, the function disables an interrupt, clears up assigned channels and stops the sys counter, if MCUboot has set these up then why should they not be cleaned up before booting an application which has it's own init functions for setting these features up in the configuration it wants (which might be completely different)? @bjarki-andreasen @nordic-krch @adamkondraciuk

ahasztag: I see you've already mentioned it in one of the conversations, missed this

PR #565, opened 2025-10-30
Modify MCUboot to use PSA key_id_t for key type in AES encryption context

de-nordic:
Two commits from upstream:
 - split of encryption headers
 - change in encryption context, where there is no longer raw key stored in encryption context, instead context setup function has been setup to import key into PSA and obtain psa_key_id_t key id for it that is then stored in context; all encryption functions will now operate on that key_id, when using the context.

Replacement of raw key array in encryption context, with psa_key_id_t, has reduced MCUboot binary size by ~224 bytes, for non-KMU and non-log-dbg build; small amount of RAM is saved also.

### DIFF
--- a/boot/bootutil/include/bootutil/crypto/aes_ctr_mbedtls.h
+++ b/boot/bootutil/include/bootutil/crypto/aes_ctr_mbedtls.h
@@ -1,0 +1,58 @@
+/*
+ * This module provides a thin abstraction over some of the crypto
+ * primitives to make it easier to swap out the used crypto library.
+ *
+ * At this point, there are two choices: MCUBOOT_USE_MBED_TLS, or
+ * MCUBOOT_USE_TINYCRYPT.  It is a compile error there is not exactly
+ * one of these defined.
+ */
+
+#ifndef __BOOTUTIL_CRYPTO_AES_CTR_MBEDTLS_H_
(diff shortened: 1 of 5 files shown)


tomchy (approved):

boot/bootutil/include/bootutil/crypto/aes_ctr_mbedtls.h
```diff
@@ -0,0 +1,58 @@
...
+{
+    mbedtls_aes_free(ctx);
+}
+
+static inline int bootutil_aes_ctr_set_key(bootutil_aes_ctr_context *ctx, const uint8_t *k)
+{
+    return mbedtls_aes_setkey_enc(ctx, k, BOOT_ENC_KEY_SIZE * 8);
+}
+
+static inline int bootutil_aes_ctr_encrypt(bootutil_aes_ctr_context *ctx, uint8_t *counter, const uint8_t *m, uint32_t mlen, size_t blk_off, uint8_t *c)
```
tomchy: Ideally, I'd prefer to have docstrings inside the interface header, but since this is a [fromtree] commit, I'm not going to require them at this stage.

nvlsianpu (approved):
reviewed upstream

PR #594, opened 2026-01-29
Code changes required to support PM removal

de-nordic:
Changes that allow to use PM or DTS defined partitions.


boot/bootutil/src/loader.c
```diff
@@ -1070,24 +1064,18 @@ boot_validated_swap_type(struct boot_loader_state *state,
                 return BOOT_SWAP_TYPE_FAIL;
```
nordicjm: code above should get reset address from load address in header, if it is 0 then it should read it from those bytes, because those bytes might be encrypted or whatnot, see https://github.com/nordicjm/sdk-mcuboot/blob/blpmdeprecation/boot/bootutil/src/loader.c#L1028

boot/bootutil/src/swap_priv.h
```diff
@@ -130,12 +130,9 @@ bool swap_write_block_size_check(struct boot_loader_state *state);
  */
 int app_max_size(struct boot_loader_state *state);
 
-#if defined(PM_S1_ADDRESS) && !defined(MCUBOOT_OVERWRITE_ONLY) && \
-CONFIG_MCUBOOT_MCUBOOT_IMAGE_NUMBER != -1
 /**
  * Performs an NSIB update
  */
 void nsib_swap_run(struct boot_loader_state *state, struct boot_status *bs);
```
tomchy: Shouldn't that be guarded by `MCUBOOT_IS_SECOND_STAGE`?

boot/zephyr/include/sysflash/sysflash.h
```diff
@@ -71,4 +178,19 @@ static inline uint32_t __flash_area_ids_for_slot(int img, int slot)
...
+#define FPROTECT_REGION_OFFSET  FIXED_PARTITION_OFFSET(boot_partition)
+#define FPROTECT_REGION_SIZE    FIXED_PARTITION_SIZE(boot_partition)
+#endif
+
+/* RWX protection regions, MCUboot is protecting itself */
+#if CONFIG_NCS_MCUBOOT_DISABLE_SELF_RWX
+#define PROTECTED_REGION_START  FIXED_PARTITION_OFFSET(boot_partition)
+#define PROTECTED_REGION_SIZE   FIXED_PARTITION_SIZE(boot_partition)
+#endif /* CONFIG_NCS_MCUBOOT_DISABLE_SELF_RWX */
+#endif
```
tomchy: I do not understand the logic here - all cases uses `boot_partition` for offset and size values, yet the comment is different. If addresses are the same - what's the difference then?

boot/zephyr/include/sysflash/sysflash.h
```diff
@@ -26,37 +43,127 @@
...
-#define ALL_AVAILABLE_SLOTS FLASH_AREA_IMAGE_0_SLOTS
-#elif (MCUBOOT_IMAGE_NUMBER == 2)
-#define ALL_AVAILABLE_SLOTS FLASH_AREA_IMAGE_0_SLOTS, \
-                            FLASH_AREA_IMAGE_1_SLOTS
-#elif (MCUBOOT_IMAGE_NUMBER == 3)
-#define ALL_AVAILABLE_SLOTS FLASH_AREA_IMAGE_0_SLOTS, \
-                            FLASH_AREA_IMAGE_1_SLOTS, \
-                            FLASH_AREA_IMAGE_2_SLOTS
+#if MCUBOOT_IS_SECOND_STAGE
+/* Defined when building MCUboot for S1 */
```
tomchy: Only `S1` or `S0` and `S1`?
If you mean the `second stage bootloader`, please use this phrase instead.

boot/zephyr/include/sysflash/sysflash.h
```diff
@@ -26,37 +43,127 @@
...
+#endif
+
+/* RWX protection regions: the currently executing MCUboot is protecting itself */
+#if CONFIG_NCS_MCUBOOT_DISABLE_SELF_RWX
+/* Note: The CONFIG_ROM_START_OFFSET is removed to save RWX bits on space
+ * that do not require protection. When MCUboot is Second Stage bootloader
+ * it does have header added, but it is not used by the NSIB.
+ */
+#define PROTECTED_REGION_START \
+    (SECOND_STAGE_ACTIVE_MCUBOOT_OFFSET + CONFIG_ROM_START_OFFSET)
```
tomchy: Why there is `CONFIG_ROM_START_OFFSET` here, while `PROTECTED_REGION_START_SKIP` is used elsewhere?

tomchy (changes requested):

boot/bootutil/src/swap_nsib.c
```diff
@@ -56,7 +54,7 @@ void nsib_swap_run(struct boot_loader_state *state, struct boot_status *bs)
     rc = boot_erase_region(fap_pri, 0, fap_pri->fa_size, false);
     assert(rc == 0);
 
-    rc = boot_copy_region(state, fap_sec, fap_pri, 0, 0, fap_pri->fa_size);
+    rc = boot_copy_region(state, fap_sec, fap_pri, hdr->ih_hdr_size, 0, hdr->ih_img_size);
```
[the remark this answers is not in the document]
de-nordic: Removed the incompatibility

boot/zephyr/Kconfig
```diff
@@ -527,6 +527,20 @@ config NCS_MCUBOOT_DISABLE_SELF_RWX
...
+	depends on !PARTITION_MANAGER_ENABLED
+	default 0x800 if SOC_SERIES_NRF54L
+	help
+	  Size of region, at the beginning of the partition, that does not require
+	  protections. This is used, for example, when MCUboot type image is protected
+	  but the header does not have to be. Allows to spare some protection bits.
+	  Note that when Partition Manager is enabled this value will be 0 and is useless.
+	  In case of images built for MCUboot the value here will be equivalent to
+	  ROM_START_OFFSET for application build for MCUboot, and it is not available
+	  in bootloader builds.
```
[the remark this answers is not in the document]
de-nordic: Previously this was only visible in code, before being exported as Kconfig.
As far as I do understand reasoning behind original implementation, the skip has been used to not protect parts that we do not care and save rwx entries for things we want to protect.
Images singed by imgtool and used for DFU and MCUboot updates will have MCUboot header added to them for MCUmgr to recognize them and for MCUboot in s0 or s1 to be able to swap the image from the DFU region to the executable region; the header is only used/recognized by MCUboot and MCUmgr. The image, nevertheless, is copied to target slot, s0 or s1, with the header.
NSIB is responsible for booting whatever is in s0 or s1, and it will skip over verification of the header, is it does not know it, and execute code starting past the header, on the original image in s0 or s1.
So we do not care about contents of range the header is in s0 or s1, that is what the skip is skipping.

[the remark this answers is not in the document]
nordicjm: ~~Query: doesn't MCUboot actually require a valid primary header in all slots to boot? I.e. if you corrupt this sector in both s0 and s1 images, whilst the code on the device should be fine, won't MCUboot fail startup then refuse to boot?~~ seems it doesn't care

boot/bootutil/src/swap_priv.h
de-nordic: It can, but honestly I have lost understanding for ifdefing function prototypes; they just make interface files look worse.

boot/zephyr/include/sysflash/sysflash.h
de-nordic: These are for different parts of code. You can change one without affecting the other.

boot/zephyr/include/sysflash/sysflash.h
```diff
@@ -26,37 +43,128 @@
...
+/* Defined when building MCUboot for S1 */
+/* We should be able to figure out which partition MCUboot is running from by
+ * comparing zephyr,code-partition to either of recognized mcuboot_sx image.
+ * It becomes problematic when variant image logic is used, because information
+ * from DT will be the same for both images, unless overlay that changes this
+ * is somehow applied; which does not happen currently.
+ * We are left with comparing CONFIG_FLASH_LOAD_OFFSET against offset of a partition.
+ * This will not work with external memory, so to be save comparison is done
+ * against both partitions each time, to find situation when we can not find
+ * a match.
```
tomchy: I see that the code below uses the `zephyr,code-partition`, so this comment seems to be obsolete?

tomchy (approved):

boot/zephyr/include/sysflash/sysflash.h
de-nordic: Yes.

nordicjm (approved):

PR #606, opened 2026-02-18
[nrf noup] boot: zephyr: cleanup of SPIM for nRF91

diegosolano:
## Summary

- Stop SPIM EasyDMA and disable all instances in `nrf_cleanup_peripheral()` before chain-loading the application
- On nRF91 with TF-M and external SPI flash, MCUboot leaves SPIM active after reading the secondary slot. When TF-M reconfigures SPU SRAM permissions, the stale DMA triggers RAMACCERR

This follows the same pattern as PR #602 (SQSPI cleanup for nRF54L) and the existing UARTE/RTC cleanup code.

## Test plan

- Tested on custom nRF9151 board with external SPI flash (MCUboot secondary slot)
- MCUboot + TF-M + non-secure application boots cleanly with this patch
- Without it, SPU fires RAMACCERR on every boot

## Context

`nrf_cleanup.c` handles peripheral teardown for peripherals that Zephyr cannot deinitialize. SPIM was missing from the cleanup list, which is a problem for any nRF91 board that uses SPI flash for the MCUboot secondary slot.

### DIFF
--- a/boot/zephyr/nrf_cleanup.c
+++ b/boot/zephyr/nrf_cleanup.c
@@ -26,7 +26,9 @@
     #include <hal/nrf_vpr.h>
     #include <softperipheral_regif.h>
 #endif
-
+#if defined(SPIM_PRESENT)
+    #include <hal/nrf_spim.h>
+#endif
 
 #include <string.h>
 
@@ -197,6 +199,33 @@ void nrf_cleanup_peripheral(void)
         memset((uint8_t *)current + NRF_UARTE_PUBLISH_CONF_OFFS, 0,
                NRF_UARTE_PUBLISH_CONF_SIZE);
 #endif
+    }
+#endif
+
+#if defined(SPIM_PRESENT)
+    /* Stop SPIM EasyDMA and disable all instances to prevent stale DMA
+     * from causing RAMACCERR when TF-M reconfigures SPU permissions. */
+    {
+        NRF_SPIM_Type *spim_inst[] = {
+#if defined(NRF_SPIM0)
+            NRF_SPIM0,
+#endif
+#if defined(NRF_SPIM1)
+            NRF_SPIM1,
+#endif
+#if defined(NRF_SPIM2)
+            NRF_SPIM2,
+#endif
+#if defined(NRF_SPIM3)
+            NRF_SPIM3,
+#endif
+        };
+
+        for (int i = 0; i < sizeof(spim_inst) / sizeof(spim_inst[0]); ++i) {
+            nrf_spim_task_trigger(spim_inst[i], NRF_SPIM_TASK_STOP);
+            nrf_spim_disable(spim_inst[i]);
+            nrf_spim_int_disable(spim_inst[i], 0xFFFFFFFF);
+        }
     }
 #endif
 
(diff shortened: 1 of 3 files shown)


ahasztag (approved):

ahasztag (approved):

boot/zephyr/nrf_cleanup.c
```diff
@@ -200,6 +202,33 @@ void nrf_cleanup_peripheral(void)
...
+#if defined(NRF_SPIM2)
+            NRF_SPIM2,
+#endif
+#if defined(NRF_SPIM3)
+            NRF_SPIM3,
+#endif
+        };
+
+        for (int i = 0; i < sizeof(spim_inst) / sizeof(spim_inst[0]); ++i) {
+            nrf_spim_task_trigger(spim_inst[i], NRF_SPIM_TASK_STOP);
```
nika-nordic: in theory code should wait for STOPPED event. Same goes for existing UARTE code. Reference:
- https://docs.nordicsemi.com/bundle/ps_nrf9160/page/spim.html#ariaid-title6
- https://docs.nordicsemi.com/bundle/ps_nrf9160/page/uarte.html#ariaid-title8

> The STOP task may not be always needed (the peripheral might already be stopped), but if it is sent, software shall wait until the STOPPED event was received as a response before disabling the peripheral through the ENABLE register.

FrancescoSer (approved):

de-nordic (changes requested):
Where is sdk-nrf manifest update that takes the change in?

diegosolano:
@de-nordic the companion manifest PR is live and rebased: https://github.com/nrfconnect/sdk-nrf/pull/27737 — CLA cleared, no conflicts. Ready for the sdk-nrf side to proceed once this one merges.

diegosolano:
@de-nordic friendly ping - the companion sdk-nrf manifest PR is now open and rebased: https://github.com/nrfconnect/sdk-nrf/pull/27737 (CLA cleared). Could you re-review here when you have a moment? Once this lands I'll bump the manifest revision to the final merge SHA. Thanks!

diegosolano:
@nika-nordic Thanks — you're right per the SPIM/UARTE spec. The Nordic docs explicitly say "if [STOP] is sent, software shall wait until the STOPPED event was received as a response before disabling the peripheral through the ENABLE register."

I've left this PR as-is for now to match the existing UARTE cleanup convention in the same file (which also issues STOP without waiting for STOPPED before `nrfy_uarte_disable()`). Doing it for only SPIM would create an inconsistency, and doing it for both feels like it belongs in its own change so the wait-for-STOPPED semantics — including a sensible timeout / fallback if the event never asserts — get reviewed on their own merits across all the peripherals in `nrf_cleanup_peripheral()`.

Happy to file that as a follow-up PR if you'd like, or to roll it into this one if you'd prefer it not ship without the fix. Let me know which you'd rather.

diegosolano:
@de-nordic the companion sdk-nrf manifest update is https://github.com/nrfconnect/sdk-nrf/pull/27737 — it's been open since Mar 26, CLA cleared, and currently points at this PR's branch tip (`488e045e`); I'll bump it to the merge SHA once this lands. The `handle manifest PR` check originally failed before #27737 existed; could a maintainer re-run that workflow now that the companion is in place? (I tried `rerun-failed-jobs` myself but it needs admin rights on the repo.) Let me know if there's anything else you need from my side here.

PR #621, opened 2026-03-12
bootutil: Add LCS handling

tomchy:
Add logic that interprets the LCS value and changes the internal boot logic, allowing to implement a provisioning application.

Ref: NCSDK-37035

### DIFF
--- a/boot/bootutil/src/ed25519_psa_kmu_its.c
+++ b/boot/bootutil/src/ed25519_psa_kmu_its.c
@@ -17,6 +17,9 @@
 #if defined(CONFIG_BOOT_SIGNATURE_USING_KMU)
 #include <cracen_psa_kmu.h>
 #endif
+#ifdef CONFIG_NCS_MCUBOOT_LCS_AWARE
+#include <nrf_lcs/nrf_lcs.h>
+#endif
 
 BOOT_LOG_MODULE_REGISTER(ed25519_psa_kmu_its);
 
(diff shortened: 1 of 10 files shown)


michalek-no (approved):

boot/zephyr/Kconfig
```diff
@@ -1532,4 +1532,20 @@ config PM_APP_ALIGNMENT
 	default 0x1000 if SOC_SERIES_NRF54L
 	default FPROTECT_BLOCK_SIZE
 
+config NCS_MCUBOOT_LCS_AWARE
+	bool "Lifecycle state awareness"
+	depends on NRF_LCS
```
nvlsianpu: Behavior implemented is for MCUboot as 2nd stage bootloader.
```suggestion
	depends on NRF_LCS && (MCUBOOT_MCUBOOT_IMAGE_NUMBER != -1)
```

boot/bootutil/src/ed25519_psa_kmu_its.c
```diff
@@ -77,23 +88,44 @@ int ED25519_verify(const uint8_t *message, size_t message_len,
...
-    for (int i = 0; i < KEY_SLOTS_COUNT; ++i) {
-        psa_key_id_t kid = key_ids[i];
+#ifdef CONFIG_NCS_MCUBOOT_LCS_AWARE
+    if (nrf_lcs_get() == NRF_LCS_SECURED) {
+#endif /* CONFIG_NCS_MCUBOOT_LCS_AWARE */
+        for (int i = 0; i < KEY_SLOTS_COUNT; ++i) {
+            psa_key_id_t kid = key_ids[i];
+
+#ifdef CONFIG_NCS_MCUBOOT_LCS_AWARE
+            if (kid == manufacturing_app_key_id) {
```
nvlsianpu: so we assume that manufacturing key might be one of keys_ids[] or be independent value.
This imply that revocation might destroys that key or not.

Maybe worth to force CONFIG_NCS_MCUBOOT_MANUFACTURING_APP_KEY_IDENTIFIER to be out of KMU slot used. What do you think?

tomchy: They are defined separately. The manufacturing key is not part of the key_ids[]. This part is to enforce that even if someone wants to, it is impossible to use manufacturing key for a regular app. I do not know how to force it differently..

boot/zephyr/Kconfig
tomchy: Is it? Personally I do not see a part that depends on the NSIB existence.

boot/bootutil/src/ed25519_psa_kmu_its.c
tomchy: Replaced with BUILD_ASSERTs

nvlsianpu (approved):

PR #640, opened 2026-04-07
[nrf noup] boot: zephyr: Support 4 images with MPCCONF write protection

jonathannilsen:
manifest-pr-skip

Update the MPCCONF tables used to write protect the image partitions
on nRF54H20 to support up to 4 images.

This is the maximum number of distinct partitions possible to write
protect with the available overrrides in MPC110 without introducing
additional restrictions, like disallowing some gaps between partitions
or adding owner ID based restrictions.

### DIFF
--- a/boot/zephyr/load_ironside_se_conf_mpc.c
+++ b/boot/zephyr/load_ironside_se_conf_mpc.c
@@ -16,41 +16,15 @@
 
 BOOT_LOG_MODULE_DECLARE(mcuboot);
 
-BUILD_ASSERT(MCUBOOT_IMAGE_NUMBER <= 2,
+BUILD_ASSERT(MCUBOOT_IMAGE_NUMBER <= 4,
 	     "The MPC override configuration used to provide write protection for the image "
-	     "partitions does not currently support more than two images.");
+	     "partitions does not currently support more than four images.");
 
 /* Required alignment for OVERRIDE addresses in MPC110. */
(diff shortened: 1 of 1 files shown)


jonathannilsen:
I have checked the 1, 2, and 3 image cases manually (for the 3 image case with a few different generated memory map permutations), but the 4 image case is not verified yet. The build time logic is getting complicated enough that it seems to deserve a dedicated test, so I will look into that.

tomchy (approved):

jonathannilsen:
> I have checked the 1, 2, and 3 image cases manually (for the 3 image case with a few different generated memory map permutations), but the 4 image case is not verified yet. The build time logic is getting complicated enough that it seems to deserve a dedicated test, so I will look into that.

From what I could tell 4 images might not be supported in NCS at the moment (missing https://github.com/mcu-tools/mcuboot/commit/3af40a3155493133ed6fe0e5252d3ef9efdf75aa ?), however I was able to do a basic test by updating sysflash.h locally and adding a 4th dummy image to nrf/samples/nrf54h20/idle_relocated_tcm, and it seemed to work as intended. So at least this feature should hopefully be future proof for if/when that gets expanded

57300 (approved):
Checked 442 configurations in total, for N = 1..4 images:

* 2^N sets of partitions with and without trailing gaps: PROTECTEDMEM and (N - 1) images (no gap after last image)
* N! slot number permutations of image partitions

PR #650, opened 2026-04-16
[nrf noup] boot: zephyr: Add mcuboot support for nRF7120

erdemsimsek:


boot/zephyr/sample.yaml
```diff
@@ -36,6 +36,7 @@ tests:
       - twr_kv58f220m
       - vmu_rt1170/mimxrt1176/cm7
       - nrf52840dk/nrf52840
+      - nrf7120dk/nrf7120/cpuapp
```
nvlsianpu: you can remove change in that file. This file is not used by sdk-mcuboot. It is just inherited from the upsteram.

adsz-nordic (approved):
Please apply Andrzej's suggestions.

boot/zephyr/sample.yaml
erdemsimsek: Addressed

nvlsianpu (approved):

PR #680, opened 2026-05-29
[nrf noup] boot/bootutil/loader: image discovery by ih_load_address

nvlsianpu:
Align all image's matching exercises to MCUBOOT_CHECK_HEADER_LOAD_ADDRES. This method uses ih_load_address field of the image header instead of reset vector address. This allows to match
incoming image to the partition even when it is for instance encrypted, as the image header is always plain-text.

Firmware need to be signed with
imgtool.py sign --rom-fixed <partition_address> parameter in order to involve this feature.

Worth to emphasize that MCUBOOT_VERIFY_IMG_ADDRESS feature was deprecated and MCUBOOT_CHECK_HEADER_LOAD_ADDRES feature is it replacement.

ref.: NCSIDB-1173, NCSIDB-1783

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -147,7 +147,7 @@ static const struct image_version mcuboot_s0_s1_image_version = {
 #endif
 
 /* Valid only for ARM Cortext M */
-#define RESET_OFFSET    (2 * sizeof(uint32_t))
+#define RESET_OFFSET sizeof(uint32_t)
 
 #if BOOT_MAX_ALIGN > 1024
 #define BUF_SZ BOOT_MAX_ALIGN
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1043,30 +1043,38 @@ boot_validated_swap_type(struct boot_loader_state *state,
 #if defined(MCUBOOT_IS_SECOND_STAGE) || CONFIG_MCUBOOT_NETWORK_CORE_IMAGE_NUMBER != -1
     const struct flash_area *secondary_fa = BOOT_IMG_AREA(state, BOOT_SLOT_SECONDARY);
     struct image_header *hdr = boot_img_hdr(state, BOOT_SLOT_SECONDARY);
-    uint32_t reset_addr = 0;
+    uint32_t internal_img_addr = 0;
     int rc = 0;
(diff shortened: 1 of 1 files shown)


michalek-no (approved):

boot/bootutil/src/loader.c
```diff
@@ -1043,43 +1043,49 @@ boot_validated_swap_type(struct boot_loader_state *state,
...
-            rc = flash_area_open(flash_area_id_from_multi_image_slot(
-                    BOOT_CURR_IMG(state),
-                    BOOT_SLOT_PRIMARY),
-                &primary_fa);
-            if (rc != 0) {
-                return BOOT_SWAP_TYPE_FAIL;
-            }
+            const struct flash_area *primary_fa =
+                BOOT_IMG_AREA(state, BOOT_SLOT_PRIMARY);
+            const uint32_t pri_off = flash_area_get_off(primary_fa);
```
tomchy, line 1089: Are you sure that this works without the `flash_area_open(..)` call?

I see that the open API calculates offsets and sizes.

boot/bootutil/src/loader.c
```diff
@@ -1088,21 +1094,21 @@ boot_validated_swap_type(struct boot_loader_state *state,
...
 
                 NSIB_OWNED_SET(BOOT_CURR_IMG(state));
-            } else if (reset_addr >= SECOND_STAGE_ACTIVE_MCUBOOT_OFFSET &&
-                       reset_addr <= (SECOND_STAGE_ACTIVE_MCUBOOT_OFFSET + SECOND_STAGE_ACTIVE_MCUBOOT_SIZE)) {
+            } else if (internal_img_addr >= SECOND_STAGE_ACTIVE_MCUBOOT_OFFSET &&
+                       internal_img_addr < (SECOND_STAGE_ACTIVE_MCUBOOT_OFFSET +
+                                            SECOND_STAGE_ACTIVE_MCUBOOT_SIZE)) {
                 /* NSIB upgrade but for the wrong slot, must be erased */
                 BOOT_LOG_ERR("Image in slot is for wrong s0/s1 image");
-                flash_area_erase(secondary_fa, 0, secondary_fa->fa_size);
```
tomchy, line 1095: Is there some other place that will erase this region or do we change the behavior here?

tomchy (approved):

de-nordic (approved):

PR #688, opened 2026-06-12
[nrf noup] bootloader: Add RSA FW encryption support

frkv:
-Add encrypted_psa.c to build when RSA encryption and signature is
 used to get access to APIs used to encrypt/decrypt images using
 AES CTR.
-Wrap boot_decrypt_key in encrypted_psa.c because it is only
 usable when ECDH using secp256r1 or X25519 is used.
-Add Kconfig for selecting RSA encryption dependencies when
 BOOT_ENCRYPT_IMAGE and RSA signatures are enabled.


de-nordic (approved):

tomchy (approved):

tomi-font (commented):
LGTM

PR #700, opened 2026-06-29
[nrf noup] bootutil: set RSA key bits for PSA OAEP decrypt

nvlsianpu:
PSA Core Lite does not infer key size from imported DER. Set psa_set_key_bits() before psa_import_key() so bootutil_rsa_oaep_decrypt accepts RSA-2048 ciphertext when validating encrypted firmware images.

skip-manifest-pr


michalek-no (approved):

frkv (approved):
LGTM

PR #702, opened 2026-07-01
Update mcuboot configuration for nrf7120

erdemsimsek:


nvlsianpu:
@michalek-no Can you check whether rework of the self RWX cut-off implementation gives proper eqivalent?

boot/zephyr/main.c
```diff
@@ -276,6 +345,15 @@ static void __ramfunc jump_in(struct arm_vector_table *vt)
     }
 #endif
 
+#if defined(CONFIG_NCS_MCUBOOT_DISABLE_SELF_RWX)
+        /* Lock MCUboot's own NVM here, from RAM and while still on the MCUboot
+         * stack (i.e. before any RAM cleanup). Everything executed afterwards
+         * lives in RAM (__ramfunc), so the now non-readable/executable MCUboot
+         * slot is never fetched or read again.
+         */
+        disable_self_rwx();
```
nvlsianpu, line 353: new procedure should works - it now uses stack, although it is moved before stack reconfiguration.
Original procedure was sackless - so it was not the case.

boot/zephyr/main.c
```diff
@@ -178,49 +178,118 @@ K_SEM_DEFINE(boot_log_sem, 1, 1);
...
+	volatile uint32_t *address = &NVM_REGION[region].ADDRESS;
+	uint32_t cfg = *config;
+
+	/* If a previous stage (NSIB) already provisioned the region the SIZE
+	 * field is non-zero; otherwise set the address and size ourselves. The
+	 * address register is left untouched when already configured, as the
+	 * region may already be locked.
+	 */
+	if ((cfg & NVM_CONFIG_SIZE_Msk) == 0) {
+		cfg &= ~NVM_CONFIG_SIZE_Msk;
```
michalek-no: clearing if it's clear?

boot/zephyr/main.c
erdemsimsek: Correct. The C implementation can use the stack because it runs before MSP replacement and RAM cleanup.

boot/zephyr/main.c
erdemsimsek: Correct, the clear was redundant because the condition already guarantees that the SIZE field is zero. Removed it.

boot/zephyr/main.c
```diff
@@ -178,49 +178,117 @@ K_SEM_DEFINE(boot_log_sem, 1, 1);
...
+/* Lock the whole MCUboot slot, chaining consecutive regions when a single one
+ * cannot cover it (nRF71). nRF54L fits in one region and loops once.
+ */
+static void __ramfunc disable_self_rwx(void)
+{
+	uint32_t remaining_kb = PROTECTED_REGION_SIZE / NVM_REGION_SIZE_UNIT;
+	uint32_t start = PROTECTED_REGION_START;
+	uint32_t region = NVM_SELF_REGION_FIRST;
+
+	while (remaining_kb) {
```
nvlsianpu, line 282: discussed off-channel witch @michalek-no Remove combined regions from that code, and use it under fprotect API instead.
Rationale: Extending this would require alignment for other S0C as well - this will imply prolonged discussions.

travis3630 (approved):

PR #704, opened 2026-07-07
boot: zephyr: decompression: reflect pad value

michalek-no:


tomchy (approved):

de-nordic:
I am not sure this is proper solution. Looks like a hack where we are trying to fill in information that has not correctly been applied to a header by copying from other header.
The proper way should be to parametrize imgtool with pad value. We are giving header size to imgtool anyway, so we, as well, request to fill it with value we want.

de-nordic (approved):

PR #727, opened 2026-08-10
[nrf noup] scripts: imgtool: padding fixes

michalek-no:
adjust to take into account compression PM-dts
compatibility

### DIFF
--- a/scripts/imgtool/image.py
+++ b/scripts/imgtool/image.py
@@ -459,7 +459,7 @@ class Image:
         except FileNotFoundError:
             raise click.UsageError("Input file not found")
 
-        self._apply_header_padding()
+        self._apply_header_padding(payload_reserves_header=not self.pad_header)
 
         self.image_size = len(self.payload) - self.header_size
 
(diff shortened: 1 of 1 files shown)


jnsgsr (approved):

scripts/imgtool/image.py
```diff
@@ -508,28 +508,34 @@ def save(self, path, hex_addr=None):
...
                 f.write(self.payload)
 
-    def _apply_header_padding(self, default_zero_fill=False):
-        """Apply header padding using the configured pad value."""
+    def _apply_header_padding(self, payload_reserves_header):
+        """Apply header padding using the configured pad value.
+
+        payload_reserves_header states whether the payload already keeps
+        header_size bytes free at its start. A binary built with
+        CONFIG_ROM_START_OFFSET does, a compressed payload never does.
```
de-nordic: Kconfigs from other system should not be mentioned in documentation.
You are also not able to ensure the information here internally,  as the reason for `payload_reserves_header` to be set changes outside of the logic within the function here.

de-nordic (changes requested):

de-nordic (approved):

PR #730, opened 2026-08-20
handling data as image

michalek-no:
adds "continue with invalid images"
meant to be used with application data handled as image

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1087,6 +1087,117 @@ static inline void sec_slot_cleanup_if_unusable(void)
 #endif /* defined(CONFIG_MCUBOOT_CLEANUP_UNUSABLE_SECONDARY) &&\
           defined(MCUBOOT_IS_SECOND_STAGE) || defined(CONFIG_SOC_NRF5340_CPUAPP) */
 
+#if defined(CONFIG_BOOT_OPTIONAL_IMAGES)
+/**
+ * Check whether an image may be invalid without blocking the boot.
+ *
+ * Image 0 is the image that MCUboot boots, so it is rejected here regardless of
+ * the configured mask.
+ *
+ * @param image_index Index of the image to check.
+ *
+ * @return true if the image is optional, false otherwise.
+ */
+static bool boot_image_is_optional(uint8_t image_index)
+{
+    if (image_index == 0) {
+        return false;
+    }
+
+    return (CONFIG_BOOT_OPTIONAL_IMAGES_MASK & (1UL << image_index)) != 0;
+}
+
+#if defined(CONFIG_BOOT_OPTIONAL_IMAGES_SCRUB)
+/**
+ * Check whether a location in a flash area still holds data.
+ *
+ * @param fap Flash area to inspect.
+ * @param off Offset within the flash area.
+ *
+ * @return true if the location is not in the erased state.
+ */
+static bool boot_region_has_data(const struct flash_area *fap, uint32_t off)
+{
+    uint32_t erased_word;
+    uint32_t first_word;
+
+    if (flash_area_read(fap, off, &first_word, sizeof(first_word)) != 0) {
+        /* The state cannot be determined, so assume a scrub is still needed. */
+        return true;
+    }
+
+    memset(&erased_word, flash_area_erased_val(fap), sizeof(erased_word));
+
+    return first_word != erased_word;
+}
+
+/**
+ * Scrub the primary slot of an optional image that failed validation.
+ *
+ * Only the first sector is scrubbed. Dropping the image header is what makes
+ * the application treat the partition as empty, and limiting the write to a
+ * single sector keeps the boot short for slots that may be large.
+ *
+ * The outcome is reported here rather than by the caller, so that a boot that
+ * finds the slot already scrubbed is not reported as having scrubbed it again.
+ *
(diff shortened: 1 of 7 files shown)


boot/zephyr/Kconfig
```diff
@@ -634,6 +612,41 @@ config BOOT_VALIDATE_SLOT0_ONCE
 	  low end devices with as a compromise lowering the security level.
 	  If unsure, leave at the default value.
 
+config BOOT_OPTIONAL_IMAGES
+	bool "Boot the application even if an optional image is invalid"
```
nvlsianpu, line 616: add ad the prompt end [EXPERIMENTAL]

boot/zephyr/Kconfig
```diff
@@ -634,6 +612,41 @@ config BOOT_VALIDATE_SLOT0_ONCE
...
+	depends on UPDATEABLE_IMAGE_NUMBER > 1
```
nvlsianpu, line 617:
```suggestion
	depends on UPDATEABLE_IMAGE_NUMBER > 1
	select EXPERIMENTAL
```

boot/bootutil/src/loader.c
```diff
@@ -1087,6 +1087,117 @@ static inline void sec_slot_cleanup_if_unusable(void)
...
+    }
+
+    rc = boot_scramble_region(fap, off, size, false);
+    if (rc != 0) {
+        BOOT_LOG_ERR("Failed to scrub image %d: %d", BOOT_CURR_IMG(state), rc);
+    } else {
+        BOOT_LOG_INF("Scrubbed image %d in the primary slot", BOOT_CURR_IMG(state));
+    }
+}
+#endif /* CONFIG_BOOT_OPTIONAL_IMAGES_SCRUB */
```
nvlsianpu, line 1183: Can these code for supporting CONFIG_BOOT_OPTIONAL_IMAGES_SCRUB be moved to separate C file?

loader.c might hold only `boot_handle_invalid_optional_image()`

nvlsianpu (changes requested):

mbarchie (commented):
It looks good considering what we need for updateable model images, I'll test it with our PoC when this is finished.

Just wondering - could we (as a future step) take this further for optional images with aliased primary/secondary slots (single physical region, no real secondary to swap/revert with)? Any leftover trailer state there is meaningless, but `img_mgmt` still calls `boot_swap_type_multi()` directly to report image state to DFU clients, so it keeps reporting pending/test and can make a mobile DFU client loop on reset. Would it make sense for mcuboot to resolve that itself for optional images (e.g. `boot_swap_type_multi()` returning `NONE`, scrubbing the trailer, or something else)?

Our PoC currently works around this by erasing the trailer from the application side, which feels suboptimal.

boot/bootutil/src/loader.c
Szynkaa, line 677: I think it would make sense to mark data partitions as non bootable, but they will always return FIH_NO_BOOTABLE_IMAGE and cause scrub

boot/bootutil/src/optional_image.c
```diff
@@ -0,0 +1,86 @@
...
+    if (!boot_region_has_data(fap, off)) {
+        /* Already scrubbed on an earlier boot; skip the write so that the
+         * memory is not rewritten on every reset.
+         */
+        return;
+    }
+
+    rc = boot_scramble_region(fap, off, size, false);
+    if (rc != 0) {
+        BOOT_LOG_ERR("Failed to scrub image %d: %d", BOOT_CURR_IMG(state), rc);
```
Szynkaa, line 84: This file is missing BOOT_LOG_MODULE_DECLARE and failing build

michalek-no:
> It looks good considering what we need for updateable model images, I'll test it with our PoC when this is finished.
> 
> Just wondering - could we (as a future step) take this further for optional images with aliased primary/secondary slots (single physical region, no real secondary to swap/revert with)? Any leftover trailer state there is meaningless, but `img_mgmt` still calls `boot_swap_type_multi()` directly to report image state to DFU clients, so it keeps reporting pending/test and can make a mobile DFU client loop on reset. Would it make sense for mcuboot to resolve that itself for optional images (e.g. `boot_swap_type_multi()` returning `NONE`, scrubbing the trailer, or something else)?
> 
> Our PoC currently works around this by erasing the trailer from the application side, which feels suboptimal.

yes

boot/bootutil/src/loader.c
michalek-no: what is the point?

boot/bootutil/src/loader.c
michalek-no: done

boot/zephyr/Kconfig
michalek-no: done

boot/zephyr/Kconfig
michalek-no: done

nvlsianpu (approved):

mbarchie (approved):

boot/bootutil/src/optional_image.c
```diff
@@ -0,0 +1,88 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
```
nvlsianpu, line 2: license is wrong - should be Nordic 5-clauses.

nvlsianpu: corrected here: https://github.com/nrfconnect/sdk-mcuboot/pull/748

PR #738, opened 2026-08-31
[nrf noup] bootutil: loader: encrypted hdr logic

michalek-no:
skips header address check if it is encrypted

### DIFF
--- a/boot/bootutil/src/loader.c
+++ b/boot/bootutil/src/loader.c
@@ -1224,6 +1224,17 @@ boot_validated_swap_type(struct boot_loader_state *state,
         sec_slot_touch(state);
 
 #ifdef  MCUBOOT_IS_SECOND_STAGE
+        bool img_addr_usable = !IS_ENCRYPTED(hdr);
+
+#ifdef MCUBOOT_CHECK_HEADER_LOAD_ADDRESS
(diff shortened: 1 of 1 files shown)


tomchy (approved):

boot/bootutil/src/loader.c
```diff
@@ -1215,6 +1215,20 @@ boot_validated_swap_type(struct boot_loader_state *state,
...
 
 #ifdef  MCUBOOT_IS_SECOND_STAGE
+#ifdef MCUBOOT_CHECK_HEADER_LOAD_ADDRESS
+        const bool img_addr_usable = true;
+#else
+        /* internal_img_addr is the reset vector read from the image payload,
+         * which for an encrypted is garbage in this context.
+         */
+        const bool img_addr_usable = !IS_ENCRYPTED(hdr);
+#endif
```
de-nordic, line 1231: This is very weird. Generally the address in image is "usable" when image is not encrypted, but the #if suggests that is it generally usable when `MCUBOOT_CHECK_HEADER_LOAD_ADDRESS` is enabled, which is not true, as the config means that it will be checked. 
Even though logically this works, it is hard to read and comprehend why.

nvlsianpu (approved):

PR #744, opened 2026-09-03
[nrf noup] image_validate: Fix bus fault in custom TLV validation hook

jtguggedal:
Derive the slot offset passed to nrf_validate_custom_tlv_data() from the validated flash area instead of from the boot state, which is not usable at every call site. single_loader.c and the firmware loaders pass a NULL state, and single-slot builds alias the secondary slot to slot0_partition so the slot index resolves past the end of the image array. Both fault.


jtguggedal:
This should possibly also be backported to v3.4 LTS branch

nvlsianpu (approved):
LGTM
@tomchy Can you confirm that the patch is correct?

tomchy (approved):

PR #749, opened 2026-09-15
[nrf noup] boot: zephyr: nRF92 MPC override configuration

jtguggedal:
The MPC write protection tables were written for nRF54H20 and are rejected by IronSide SE on nRF92.

This patch now differentiates between SoCs in three areas:

* Override indices. For nRf9251, IronSide SE allows MPC110 overrides 4, 5, 6, 7 and 29. For now, we will only make use of 4, 7 and 29 for a single image, which a BUILD_ASSERT enforces.

* OWNERID and MASTERPORT. The owner of these overrides is fixed to the application core in hardware, so NRF_OWNER_NONE is refused.

* Secure storage placement. The top of MRAM belongs to the Cellular domain, so the end-of-MRAM assertion is replaced by one requiring only that secure storage sits at the end of the application core-owned region of MRAM.

Verified on an nRF9251 DK with IronSide SE 23.8.0+36.


boot/zephyr/load_ironside_se_conf_mpc.c
```diff
@@ -246,6 +248,12 @@ BUILD_ASSERT((PARTITION_ADDRESS(secure_storage_partition) +
 CHECK_MPC_ADDRESS_ALIGNMENT("Bootloader partition", ACCESSIBLE_MRAM_START);
 CHECK_MPC_ADDRESS_ALIGNMENT("MRAM end / secure storage", ACCESSIBLE_MRAM_END);
 CHECK_MPC_ADDRESS_ALIGNMENT("Bootloader partition", BOOT_PARTITION_END);
+
+#if defined(CONFIG_SOC_SERIES_NRF92) && \
+	DT_NODE_HAS_STATUS_OKAY(DT_NODELABEL(secure_storage_partition))
+BUILD_ASSERT(PARTITION_ADDRESS(secure_storage_partition) >= BOOT_PARTITION_END,
```
tomchy: In that case, the `ACCESSIBLE_MRAM_END` must not be based on the `secure_storage_partition` address.

tomchy (changes requested):

boot/zephyr/load_ironside_se_conf_mpc.c
jtguggedal: We are reworking this to align with how it's done for nRF54H20, with secure_storage_partition indicating end address. Will push an update soon

jtguggedal:
@tomchy I have now updated the logic to be similar to nRF54H20, and added some comments where nRF92 differs from the 54

Trond-F-Christiansen (approved):

nvlsianpu (approved):

tomchy (approved):

